PR #63, opened 2018-11-06
Do not expand hostnames which contains puncations marks.

### LINKED ISSUE #62: Ikke ekspander hostnavn om det inneholder en "."
`host info example.org` bør gjøre oppslag mot "example.org" og ikke  "example.org.domene.i.conf".

oyvindkolbu:
Might later require an explicit trailing dot, to distinguish
"sub.sub2" which should be interpreted as "sub.sub2.domain.in.conf"
from the fqdn "sub.tld".

Resolves #62.

### DIFF
--- a/util.py
+++ b/util.py
@@ -325,7 +325,11 @@ def resolve_ip(ip: str) -> str:
 
 def resolve_input_name(name: str) -> str:
     """Tries to find the named host. Raises an exception if not."""
-    hostname = clean_hostname(name.lower())
+    if "." in name:
+        hostname = name
+    else:
+        hostname = clean_hostname(name)
+
     url = "http://{}:{}/hosts/?name={}".format(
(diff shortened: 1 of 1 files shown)


util.py
```diff
@@ -325,7 +325,11 @@ def resolve_ip(ip: str) -> str:
 
 def resolve_input_name(name: str) -> str:
     """Tries to find the named host. Raises an exception if not."""
-    hostname = clean_hostname(name.lower())
+    if "." in name:
+        hostname = name
+    else:
+        hostname = clean_hostname(name)
```
oyvindhagberg: Ser at `name.lower()` har blitt til `name`, er det meningen?

oyvindkolbu: Nei, det var det jo ikke. Legges til!

Issue #102, opened 2019-03-28
SSHFP support


oyvindkolbu:
Client part of usit-gd/mreg#274.

dmykarp:
https://github.com/usit-gd/mreg-cli/pull/105

PR #105, opened 2019-04-11
Sshfp related commands

dmykarp:


dmykarp:
In this PR I've also fixed a couple of spotted typos in the comments not related to SSHFP.

mreg_cli/util.py
```diff
@@ -40,6 +40,19 @@ def host_exists(name: str) -> bool:
...
+    """
+    Return a dict with host information of the host having the given id.
+
+    :param id: id of the host in the datbase.
+    :return: A dict of the JSON object received with the host information
+    """
+    path = f"/hosts/?id={id}"
+    history.record_get(path)
+    host = get(path).json()
+    return host["results"][0]
```
oyvindkolbu: The `get` can result in a empty results list, and then the return will crash.

mreg_cli/host.py
```diff
@@ -2327,6 +2338,182 @@ def srv_show(args):
...
+        sshfps = get_list(path)
+        if len(sshfps) < 1:
+            cli_warning("no SSHFP records matching {}".format(info["name"]))
+    else:
+        # Get all SSHFPs
+        path = f"/sshfps/"
+        history.record_get(path)
+        sshfps = get_list(path)
+        if len(sshfps) < 1:
+            cli_warning("SSHFP records are not found")
```
oyvindkolbu: Hmm, is the "get all SSHFPs" needed? Think this is the only method capable of grabbing all entries for a certain model.

mreg_cli/host.py
```diff
@@ -2327,6 +2338,182 @@ def srv_show(args):
...
+    
+    padding = 0
+    prev_name = ""
+    for sshfp in sorted(sshfps, key=lambda k: k["host"]):
+        host_info = host_info_by_id(sshfp["host"])
```
oyvindkolbu: Cache/check if host was the previous one?

mreg_cli/host.py
```diff
@@ -2327,6 +2338,182 @@ def srv_show(args):
...
+        hname = host_info["name"]
+        if len(hname) > padding:
+            padding = len(hname)
```
oyvindkolbu: This will cause the padding to get longer and longer and thus look a bit weird. Possible to check names first to get proper padding?

mreg_cli/host.py
dmykarp: According to the docs all SRV entries can be listed with a single command. However, the code does not support that. I wanted to submit a bug report about it.

Shall we drop "get all entries" commands altogether? The bug report for SRV is still needed then, but for correcting the docs. I can drop support for all sshfps if we agree to drop it.

mreg_cli/util.py
dmykarp: This subrouting will be gone in the next commit, since we drop listing all sshfp records and no other code needs it.

mreg_cli/host.py
dmykarp: Agreed to drop this, it will be gone in the next commit.

oyvindkolbu: Also drop the docs for all SRVs.. for big data dumps you can use the API.

mreg_cli/host.py
dmykarp: Will be gone in the next commit as well.

mreg_cli/host.py
dmykarp: Will be gone in the next commit.

oyvindkolbu:
Resolves #102.

PR #124, opened 2020-04-17
Make IPv6 address removal case-insensitive

### LINKED ISSUE #123: Removing IPv6 address seems to not work because it's case sensitive
```
mreg> host aaaa_remove fys-gw 2001:700:100:AA0::246
WARNING: : 2001:700:100:AA0::246 is not owned by fys-gw.uio.no
mreg> host aaaa_remove fys-gw 2001:700:100:aa0::246
OK: : removed ip 2001:700:100:aa0::246 from fys-gw.uio.no
mreg>  
```
Det er da ikke forskjell på 0xAA0 og 0xaa0 ?

(rapportert inn av andreasd)

oyvindhagberg:
Fixes #123.

### DIFF
--- a/mreg_cli/host.py
+++ b/mreg_cli/host.py
@@ -1012,7 +1012,7 @@ def _ip_remove(args, ipversion):
     # Check that ip belongs to host
     info = host_info_by_name(args.name)
     for rec in info["ipaddresses"]:
-        if rec["ipaddress"].lower() == args.ip.lower():
+        if rec["ipaddress"] == args.ip.lower():
             ip_id = rec["id"]
             break
     else:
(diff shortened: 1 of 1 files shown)


mreg_cli/host.py
```diff
@@ -1012,7 +1012,7 @@ def _ip_remove(args, ipversion):
     # Check that ip belongs to host
     info = host_info_by_name(args.name)
     for rec in info["ipaddresses"]:
-        if rec["ipaddress"] == args.ip:
+        if rec["ipaddress"].lower() == args.ip.lower():
```
paalbra, line 1015: I'm pretty sure mreg itself is strictly lowercase. Maybe remove the first `lower()` since it's unnecessary?

oyvindhagberg: Added a new commit that does this.

mbakke (approved):
:shipit:

PR #140, opened 2021-01-06
Fix 138

### LINKED ISSUE #138: Getting host with # or ? as name will fail
Commands that tries to get hosts with `#` or `?` as name will fail.

If you try to get `/api/v1/hosts/#` or `/api/v1/hosts/?` you will actually get all hosts because:
```
>>> requests.compat.urljoin("https://example.com", "/api/v1/hosts/#")
'https://example.com/api/v1/hosts/'
>>> requests.compat.urljoin("https://example.com", "/api/v1/hosts/?")
'https://example.com/api/v1/hosts/'
```

https://github.com/unioslo/mreg-cli/blob/05011b15ff4b964acb3b2c12114e0df9d3920a32/mreg_cli/util.py#L277

Proper encoding of the name could be a fix. Or the given name should be considered invalid somehow.

E.g:
`host info #` and `dhcp disassoc #` will fail. Probably other commands as well?

oyvindhagberg:
Found a more elegant solution to #138, that doesn't require code changes everywhere:

- The utility function `clean_hostname` can also verify that the hostname doesn't contain any illegal/invalid characters, and raise an exception if it does. This function is used everywhere where a hostname is given as a parameter.
- Let `_host_info_by_name` encode the hostname in the path. The hostname should already have been vetted at this point, but it's nice as an extra safety measure.

Fixes #138.


mreg_cli/util.py
```diff
@@ -274,6 +275,7 @@ def result_check(result, type, url):
 
 
 def _request_wrapper(type, path, params={}, ok404=False, first=True, **data):
+    path = urllib.parse.quote(path, safe = "/")
```
paalbra, line 278: This will lead to `E251 unexpected spaces around keyword / parameter equals` if we do linting. Fix the whitespace?

paalbra:
I'm pretty sure that this is a decent fix as long as we don't produce paths with weird characters (#, ?, etc).

I think there is a possibility to pass "bad" parameters though. Like a hostname: `some/host`. This might produce weird results, but might be fine for now.

There is this one linting issue I mentioned above.

And maybe clean up the commit history? Some commits revert other.

Other than that: LGTM

mreg_cli/util.py
oyvindhagberg: Fixed

paalbra (approved):

oyvindkolbu (dismissed):
:shipit:

PR #145, opened 2021-04-09
Fix a bug: Wildcard records aren't accepted

oyvindhagberg:
The bug was introduced in https://github.com/unioslo/mreg-cli/pull/140:
https://github.com/unioslo/mreg-cli/blob/ed189364c0a4a21e769e0acf68a954855064f3d7/mreg_cli/util.py#L413-L415
Apparently I forgot that [wildcard DNS records](https://en.wikipedia.org/wiki/Wildcard_DNS_record) are a thing.


mreg_cli/util.py
```diff
@@ -411,7 +411,7 @@ def clean_hostname(name: typing.AnyStr) -> str:
     name = name.lower()
 
     # invalid characters?
-    if re.search("^([a-z0-9_][a-z0-9\-]*\.?)*$", name) is None:
+    if re.search("^(\*\.)?([a-z0-9_][a-z0-9\-]*\.?)*$", name) is None:
```
oyvindkolbu, line 414: Suggested change: `re.search("^(\*\.)?([a-z0-9_][a-z0-9\-]*\.?)+$", name)`
To force a non-empty match.

oyvindkolbu (changes requested):

mreg_cli/util.py
oyvindhagberg: Good point. Did it now in f104d7b

oyvindkolbu (approved):

PR #146, opened 2021-04-20
Feature: Hosts can have a BACnet Instance Number

oyvindhagberg:
This commit adds subcommands to the host command to assign BACnet IDs to hosts.
A BACnet ID (instance number) is an integer in the range 0-4194302, and hosts can't share the same ID.

### DIFF
--- a/mreg_cli/bacnet.py
+++ b/mreg_cli/bacnet.py
@@ -1,0 +1,105 @@
+from .cli import Flag, cli
+from .exceptions import HostNotFoundWarning
+from .history import history
+from .host import host
+from .log import cli_error, cli_info, cli_warning
+from .util import (
+    clean_hostname,
+    host_info_by_name,
+    _host_info_by_name,
+    resolve_input_name,
+    print_table,
+    get,
+    get_list,
+    post,
+    delete,
+)
+
+def bacnetid_add(args):
+    info = host_info_by_name(args.name)
+    if 'bacnetid' in info and info['bacnetid'] is not None:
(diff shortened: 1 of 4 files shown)


mreg_cli/bacnet.py
```diff
@@ -0,0 +1,105 @@
+from .cli import Flag, cli
+from .exceptions import HostNotFoundWarning
+from .history import history
+from .host import host
+from .log import cli_error, cli_info, cli_warning
+from .util import (
+    clean_hostname,
+    host_info_by_name,
+    _host_info_by_name,
+    resolve_input_name,
```
mbakke, line 10: Some of these imports are unused.

mreg_cli/host.py
```diff
@@ -459,6 +459,14 @@ def print_txt(txt: str, padding: int = 14) -> None:
     print("{1:<{0}}{2}".format(padding, "TXT:", txt))
 
 
+def print_bacnetid(bacnetid: dict, padding: int = 14) -> None:
+    """Pretty print given txt"""
+    if bacnetid is None:
+        return
+    assert isinstance(bacnetid, dict)
```
mbakke, line 466: Is this necessary, given the type annotation?

mbakke (approved):
Apart from a few minor comments, everything LGTM :shipit:

mreg_cli/bacnet.py
oyvindhagberg: Fixed now in b1ce84c1afd93f8751371d2e2355d02d6c930b7a

PR #149, opened 2021-06-30
ci: Switch to GitHub actions.

mbakke:
Travis has shut down their free tier, and GitHub actions are much more convenient anyway.

This also adds Python 3.9 to the test matrix.


mbakke:
I added a new commit that runs the test suite against the latest mreg Docker container, in addition to the fixtures.  It requires https://github.com/unioslo/mreg/pull/450.

oyvindhagberg (approved):

mreg_cli/main.py
```diff
@@ -111,7 +112,7 @@ def main():
     if 'mock_traffic' in conf:
         if 'record_traffic' in conf:
             print("You can't use both the playback and record options at the same time!")
-            raise SystemExit()
```
oyvindkolbu, line 114: The usage of `SystemExit` instead of just `exit(1)` was there for reason I can recall, but it might be an outdated reason.

oyvindkolbu (approved):

mreg_cli/main.py
mbakke: Thanks for the heads-up.  I see there is code in `cli.py` for dealing with `SystemExit` exceptions, and decided to drop this commit until I understand the implication better.  It does not matter anymore now that I got the arguments right. :stuck_out_tongue_winking_eye: 

Also, `sys.exit(foo)` is [just an alias for `raise SystemExit(foo)`](https://docs.python.org/3/library/sys.html#sys.exit), so it would be better to keep the `raise SystemExit` style.

Issue #152, opened 2021-11-03
Dead code in `host cname_show`


paalbra:
Pretty sure this is dead code after the except pass:

https://github.com/unioslo/mreg-cli/blob/5e6bda69f1b36818139a215164dc4997373484ce/mreg_cli/host.py#L1390-L1414

I think the above is intended to fallback to checking if input was a CNAME? But `host_info_by_name` follows CNAMEs.

https://github.com/unioslo/mreg-cli/blob/5e6bda69f1b36818139a215164dc4997373484ce/mreg_cli/util.py#L109

PR #154, opened 2021-12-07
Atomic CNAME updates

mbakke:
...as well as a fix for #152 while at it.

This PR adds a `cname_replace` command that can move a CNAME from one host to another in the same transaction.

The first commit is unrelated, but can simplify things in the mreg Docker container.

### DIFF
--- a/mreg_cli/host.py
+++ b/mreg_cli/host.py
@@ -1382,6 +1382,46 @@ host.add_command(
     ],
 )
 
+###################################################
+#  Implementation of sub command 'cname_replace'  #
+###################################################
+
+def cname_replace(args):
+    """Move a CNAME entry from one host to another.
+    """
+
+    cname = clean_hostname(args.cname)
+    host = clean_hostname(args.host)
+
+    cname_info = host_info_by_name(cname)
+    host_info = host_info_by_name(host)
+
+    if cname_info['id'] == host_info['id']:
+        cli_error(f"The CNAME {cname} already points to {host}.")
+
+    # Update CNAME record.
+    data = {'host': host_info['id'], 'name': cname }
+    path = f"/api/v1/cnames/{cname}"
+    history.record_patch(path, "", data, undoable=False)
+    patch(path, **data)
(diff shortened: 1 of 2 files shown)


mreg_cli/host.py
```diff
@@ -1382,6 +1382,46 @@ def cname_remove(args):
...
+    """
+
+    cname = clean_hostname(args.cname)
+    host = clean_hostname(args.host)
+
+    cname_info = host_info_by_name(cname)
+    host_info = host_info_by_name(host)
+
+    if cname_info['id'] == host_info['id']:
+        cli_error("The CNAME {} already points to {}.".format(cname, host))
```
oyvindkolbu, line 1400: f-strings? `f"The CNAME {cname} already points to {host}."`

mreg_cli/host.py
```diff
@@ -1382,6 +1382,46 @@ def cname_remove(args):
...
+
+    # Update CNAME record.
+    data = {'host': host_info['id'], 'name': cname }
+    path = "/api/v1/cnames/{}".format(cname)
```
oyvindkolbu, line 1404: f-string

mreg_cli/host.py
```diff
@@ -1382,6 +1382,46 @@ def cname_remove(args):
...
+    history.record_patch(path, "", data, undoable=False)
+    patch(path, **data)
+    cli_info("Moved CNAME alias {}: {} -> {}".format(cname, cname_info['name'], host),
```
oyvindkolbu, line 1407: f-string

oyvindkolbu (changes requested):
Looks better with f-strings

mbakke:
Changed to f-strings, thanks!

oyvindkolbu (approved):
:shipit: 

Please don't amend/squash and force push, as it make reviews harder.

mbakke:
> Please don't amend/squash and force push, as it make reviews harder.

Fair point, noted.  GitHub could easily show a diff of that force-push, but don't; GitLab does better in that regard. I'll delay my rebases until post-review in the future.  :-)

PR #156, opened 2022-08-12
Improve & Refactor Type Annotations

pederhan:
This pull requests improves or expands upon the type annotations found in `host.py`, `log.py` and  `util.py`. 

This pull request does not change any actual functionality within `mreg_cli` and is purely aimed at improving the developer experience.

Each change is documented in its own section below.

## Change `AnyStr` Annotations to `Union[str, bytes]`

`AnyStr` annotations are changed to `Union[str, bytes]` to signify that the argument can be converted from `bytes` to `str` (and vice versa) inside the function body.

### Rationale 

The way that `AnyStr` is currently used in the code is not legal and causes type checkers to complain. `AnyStr` is an alias for a generic type that can be `str` _or_ `bytes`, but is _not_ equivalent to `Union[str, bytes]`. 

`AnyStr` works roughly like this:

```py
AnyStr = TypeVar("AnyStr", str, bytes)

# Allows for:

def foo(s: AnyStr) -> AnyStr:
    if isinstance(s, str):
        s = s + "!"
        return s
    else:
        s = s + b"!"
        return s


# But does NOT allow:
def bar(s: AnyStr) -> AnyStr:
    if isinstance(str, bytes):
        s = s.decode() # s was bytes, is now str
    return s + "!" 


def baz(s: AnyStr) -> AnyStr:
    if isinstance(s, str):
        s = s.encode() # s was str, is now bytes
    return s + b"!"
```

During the variable's lifetime inside the function body, it should not change type.

Since all functions in `mreg_cli` with the `AnyStr` type annotation are manipulating the argument and changing its type, the type annotation `Union[str, bytes]` is more appropriate to allow for the variable's type to change within the given constraints (`str` or `bytes`).

### Commits

These changes are implemented in the following commits:

* 88bfa260a53ba079efb8dc11a613260c2551deeb

## Add Stubs with Overloads for Union Types

Functions that return `Union[T, S, ...]` should have `typing.overload` overloads either in the source file itself or in a separate stub file to specify the conditions that return `T` and `S`.

Example:

```py
from typing import overload, Literal, Union


@overload
def foo(s: str, as_list: Literal[True]) -> list:
    ...


@overload
def foo(s: str, as_list: Literal[False]) -> str:
    ...


def foo(s: str, as_list: bool) -> Union[str, list]:
    if as_list:
        return [c for c in s]
    return s + "!"
```

By specifying these overloads, type checkers are able to discern whether or not the function will return a `list` or `str` based on the arguments passed to the function.

### Implementation

The files `__init__.pyi`, `log.pyi`, as well as `py.typed` (see: [PEP 561](https://peps.python.org/pep-0561/)) have been added to provide additional type information for the `cli_<level>` functions in `log.py`. These functions are a bit problematic with regards to typing because they can either return `None` or raise an Exception. 

In order to communicate to the type checker that these functions sometimes don't return, their return type annotations have been set to `Optional[NoReturn]`. `NoReturn` signfies that the function never returns. However, that is only the case for these functions when `raise_exception` is `True`, and as such there is a need to declare overloaded definitions for these functions.

https://github.com/unioslo/mreg-cli/blob/8015897f74ab3f19c7a5b093b12edea0e696a5a8/mreg_cli/log.py#L61-L63
https://github.com/unioslo/mreg-cli/blob/8d980045473096598a8e93af344ae553d6376986/mreg_cli/log.pyi#L34-L45

#### Type Stubs

The overloaded function definitions are placed in a separate `log.pyi` stub file in order to not clutter the original `log.py` with these. Whether or not to create separate `.pyi` files is up to preference, and both conventions seem to be used. 

Examples:
* [more-itertools](https://github.com/more-itertools/more-itertools/tree/master/more_itertools) (Uses `.pyi` stubs)
* [pytest](https://github.com/pytest-dev/pytest/blob/58cf20edf08d84c5baf08f0566cc9bccbc4ec7fd/src/_pytest/fixtures.py#L1213-L1240) (Defined in source file)

### Commits

Stubs and `NoReturn` annotations are added in the following commits:

* 8015897f74ab3f19c7a5b093b12edea0e696a5a8
* 8d980045473096598a8e93af344ae553d6376986
* 03021bc33d94fc42f65d1d8ea17e25a0ad41b342

## Add `Optional[]` Annotations to Functions That Can Return `None`

Functions that can return `None` are annotated with `Optional[T]`. This change is only implemented for functions that already have a type annotation; no new annotations are added.

### Commits

This is added in the following commits:

* e36475280a23f9c8a8c070692232ca2e2b6e31d0

## Add `typing_extensions` Dependency

Certain `typing` classes, such as `Literal` are not available in Python \<3.8. To access these in Python 3.7 and below, the [`typing_extension`](https://pypi.org/project/typing-extensions/) package has been added as a dependency. This is a relatively small dependency (25.6 kB), and as such should not affect the build time or distribution size of `mreg_cli` notably.

### Commits

Added in:

* c3c428f33544232faa760044da928ab63e5734a4

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ install_requirements = [
     'python-dateutil',
     'prompt_toolkit>=2',
     'requests',
-    'typing_extensions'
+    'typing_extensions;python_version<"3.8"'
 ]
 
 
--- a/setup.py
+++ b/setup.py
@@ -10,6 +10,7 @@ install_requirements = [
     'python-dateutil',
     'prompt_toolkit>=2',
     'requests',
+    'typing_extensions'
 ]
 
 
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ install_requirements = [
     'python-dateutil',
     'prompt_toolkit>=2',
     'requests',
-    'typing_extensions;python_version<"3.8"'
+    'typing_extensions'
 ]
 
 
--- a/mreg_cli/__init__.pyi
+++ b/mreg_cli/__init__.pyi
@@ -1,0 +1,1 @@
+from .log import *
--- a/mreg_cli/host.py
+++ b/mreg_cli/host.py
@@ -1,5 +1,6 @@
 import ipaddress
-import typing
+from typing import Iterable, Optional
+
 
 from .cli import Flag, cli
 from .dhcp import assoc_mac_to_ip
@@ -53,7 +54,7 @@ def print_hinfo(hinfo: dict, padding: int = 14) -> None:
     print("{1:<{0}}cpu={2} os={3}".format(padding, "Hinfo:", hinfo['cpu'], hinfo['os']))
 
 
-def zoneinfo_for_hostname(host: str) -> dict:
+def zoneinfo_for_hostname(host: str) -> Optional[dict]:
     """Return zoneinfo for a hostname, or None if not found or invalid"""
     if "." not in host:
         return None
@@ -374,8 +375,9 @@ def print_comment(comment: str, padding: int = 14) -> None:
     print("{1:<{0}}{2}".format(padding, "Comment:", comment))
 
 
-def print_ipaddresses(ipaddresses: typing.Iterable[dict], names: bool = False,
-                      padding: int = 14) -> None:
+def print_ipaddresses(
+    ipaddresses: Iterable[dict], names: bool = False, padding: int = 14
+) -> None:
     """Pretty print given ip addresses"""
     def _find_padding(lst, attr):
         return max(padding, max([len(i[attr]) for i in lst])+1)
--- a/mreg_cli/log.py
+++ b/mreg_cli/log.py
@@ -2,6 +2,7 @@ import getpass
 import inspect
 import re
 from datetime import datetime
+from typing import NoReturn, Optional, Type
 
 from .exceptions import CliError, CliWarning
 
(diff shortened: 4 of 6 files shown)


oyvindhagberg (approved):
Looks good to me! Will merge unless anyone protests.

setup.py
```diff
@@ -10,6 +10,7 @@
     'python-dateutil',
     'prompt_toolkit>=2',
     'requests',
+    'typing_extensions'
```
mbakke, line 13: This should be

```
'typing_extensions' ; python_version<'3.8'
```

...to avoid pulling in the `typing_extensions` dependency on later Pythons.

Otherwise looks great, thanks! :)

mbakke (approved):

setup.py
pederhan: Good point. There are probably no typing changes from Python >=3.9 that need to be pulled in via `typing_extensions` foreseeable future from what I can tell.

pederhan: Actually, on second thoughts, I think `typing_extensions` should be pulled in regardless of Python version to avoid having to perform Python version checks when importing these annotations. 

`typing_extensions` already performs this check for us: https://github.com/python/typing_extensions/blob/fb77519b198f9d610ee8f674e27b8ce1da00f61f/src/typing_extensions.py#L228-L229

Furthermore, leaving it in for all Python versions makes it easier to write code that is compatible with all supported versions of Python, e.g:

```py
from typing_extensions import Literal, Protocol

class MyProto(Protocol):
    def foo() -> Literal["bar"]: ...
```

Instead of:

```py
import sys
if sys.version_.info < (3, 8):
    from typing_extensions import Literal, Protocol
else:
    from typing import Literal, Protocol

class MyProto(Protocol):
    def foo() -> Literal["bar"]: ...
```

Right now, the `.pyi` modules will show import errors in IDEs if pip installing `mreg_cli` on Python >=3.8, due to `typing_extensions` not being installed. This of course has no runtime implications\*, as these files are ignored by the interpreter, but it stands to reason that including such a small dependency for all versions should be a small concession to make to avoid having to write code like in the second example. 

\* Until `Literal`, `Protocol`, etc. are attempted to be imported into`.py` modules from `typing_extensions` at a later point.

mbakke: Oh, I thought the namespace was the same.  I agree that using `typing_extensions` unconditionally and consistently is easier.  LGTM!  :-)

PR #158, opened 2022-08-25
Use bytes.decode() instead of str() for byte string emails

pederhan:
Passing a byte string to `str()` will return the `__repr__`/`__str__` of the bytes object, not its UTF-8 string representation.

```pyconsole
>>> str(b"user@example.com")
"b'user@example.com'"
```

In order to parse the email correctly, it should be decoded via `.decode()`:

```pyconsole
>>> b"user@example.com".decode()
'user@example.com'
```

The default encoding is UTF-8, and that seems like a reasonable assumption to make when attempting to decode an arbitrary byte string.

### DIFF
--- a/mreg_cli/util.py
+++ b/mreg_cli/util.py
@@ -599,7 +599,7 @@ def is_valid_email(email: Union[str, bytes]) -> bool:
     """Check if email looks like a valid email"""
     if not isinstance(email, str):
         try:
-            email = str(email)
+            email = email.decode()
         except ValueError:
             return False
     return True if re.match(r"^[^\s@]+@[^\s@]+\.[^\s@]+$", email) else False


petterreinholdtsen:
[pederhan]
> -- Commit Summary --
>
>   * Use bytes.decode() instead of str() for bytes emails

For emails, is it not better to consult the Content-type MIME header
before making assumtions about their content?

--
Happy hacking
Petter Reinholdtsen

pederhan:
> For emails, is it not better to consult the Content-type MIME header before making assumtions about their content?

Absolutely, but as it stands the function only takes a single argument (the email string), so this would require a larger change. One could potentially do something like this:

```py
def is_valid_email(email: Union[str, bytes], encoding: str = "utf-8") -> bool:
    if not isinstance(email, str):
        try:
            email = email.decode(encoding=encoding)
    # ...
```

Where the `encoding` argument would only be used in case a byte string is passed to the function. This would be backwards-compatible with the current usage of the function while also ensuring the presumed encoding of the bytes string can be changed if needed.

----
**NOTE**

From looking at the code, `is_valid_email` is only used by CLI functions, so I'm not even sure it can ever receive a bytes object currently. 

https://github.com/unioslo/mreg-cli/blob/36ac74cb6728f67baca91038643e51228a40fb08/mreg_cli/host.py#L193

https://github.com/unioslo/mreg-cli/blob/36ac74cb6728f67baca91038643e51228a40fb08/mreg_cli/host.py#L759

pederhan:
FWIW, this behavior was discovered when writing tests for the function:

https://github.com/pederhan/mreg-cli/blob/2c2455b8360b2bf16147d38f12cc2af2f19ba1a4/tests/test_util.py#L834-L855

Where an invalid email passed in as bytes (`b"@example.com"`) would successfully pass validation when it shouldn't have.

This pull request makes that invalid email fail to validate as it should.

oyvindhagberg (approved):

petterreinholdtsen:
[Petter Reinholdtsen]
> For emails, is it not better to consult the Content-type MIME header
> before making assumtions about their content?

[pederhan]
> Absolutely, but as it stands the function only takes a single argument
> (the email string), so this would require a larger change. One could
> potentially do something like this:

It is worth noting that RFC 5322 specify that the charset of bodies
without Content-type value is US-ASCII, see section 2.3 of
<URL: https://www.rfc-editor.org/rfc/rfc5322 >.

The email.message library might be useful in this context,
<URL: https://docs.python.org/3/library/email.examples.html >.

-- 
Happy hacking
Petter Reinholdtsen

pederhan:
Yes, originally this was the case, but newer standards have expanded email addresses beyond the original 7-bit ASCII:

https://stackoverflow.com/a/2071250
https://www.rfc-editor.org/rfc/rfc6532#section-3.2 (proposed, but adoption is growing)
http://www.faqs.org/rfcs/rfc3490.html

Furthermore, since ASCII is a subset of UTF-8, we should safely be able to decode ASCII encoded text using `"utf-8"` as the encoding.

```py
>>> "foo".encode("ascii").decode("utf-8")
'foo'
```

Anyway, nothing calls this function with bytes arguments as of now, so it's probably better to revisit this once that's the case.

Issue #144, opened 2021-03-24
New command "network find"?


paalbra:
It's hard to get a overview of networks. It would be nice to e.g. list networks on a given vlan.

Maybe add a "network find" command? Kind of like the "host find" command.

Possible search parameters that would be nice to have: category, location, vlan. Other/more?

https://github.com/unioslo/mreg/blob/75011f1b5a3e9b80a4a15d9a666c15a39f2ee160/mreg/models.py#L481-L489

oyvindhagberg:
This was done in https://github.com/unioslo/mreg-cli/pull/159.

PR #159, opened 2022-09-07
Add `network find` command

pederhan:
This pull request adds the command `network find`, which was requested in issue #144 .

The command is able to search for networks based on all attributes available for MREG networks. Section **Parameters** lists information about the available query parameters used by this command.

## Parameters

At least one query parameter is required to run the command.

**Query Parameters**

* `-network <NETWORK>`: Complete or partial network address. Supports wildcards.
* `-description <DESCRIPTION>`: Complete or partial network description. Supports wildcards.
* `-vlan <VLAN>`: Complete or partial VLAN ID. Supports wildcards.
* `-dns_delegated <0-1>`: Only list networks with the given DNS delegation status.
* `-category <CATEGORY>`: Complete or partial category information. Supports wildcards.
* `-location <LOCATION>`: Complete or partial category information. Supports wildcards.
* `-frozen <0-1>`: Only list networks with the given frozen status.
* `-reserved <RESERVED>`: Exact number of reserved network addresses.

**Output Parameters**

* `-addr-only`: Only prints the network address of the matching networks instead of the full network information.
* `-limit <LIMIT>`: Limits the output to `LIMIT` matching networks. Networks beyond the limit are omitted.
* `-silent`: Silences printing of meta-info such as number of networks found and limit reached.

## Example Usage

**NOTE**: all examples are based on real usage, but values have been modified as to not reflect real networks.

### Find Network Based on Partial Network Address

```sh
mreg> network find -network 122.199.100.8*
Network:                 122.199.100.8/30
Netmask:                 255.255.255.252
Description:             i am a description
Category:                core
Location:
VLAN                     5
DNS delegated:           False
Frozen                   False
IP-range:                122.199.100.8 - 129.199.100.11
Reserved host addresses: 3
                         129.199.100.8 (net)
                         129.199.100.9
                         129.199.100.10
                         129.199.100.11 (broadcast)
Used addresses:          2
Unused addresses:        0 (excluding reserved adr.)

Found 1 network matching the search criteria.
```

### Limiting Output Length

Certain queries can lead to too many results to display at once. To remedy this, the length of the output can be controlled with the two parameters `-addr-only` and `-limit`.

```sh
mreg> network find -network 122.199.100.* -limit 5 -addr-only
129.199.100.8/30
129.199.100.12/30
129.199.100.16/30
129.199.100.20/30
129.199.100.24/30
Reached limit (5). Omitted 75 networks.
Found 80 networks matching the search criteria.
```

Furthermore, the messages displayed after the networks can be silenced with `-silent`.

```sh
mreg> network find -network 122.199.100.* -limit 5 -addr-only -silent
129.199.100.8/30
129.199.100.12/30
129.199.100.16/30
129.199.100.20/30
129.199.100.24/30
```

## Refactoring `network info` to Avoid Code Duplication

In order to re-use the network printing functionality defined in `network.info()`, most of the function's contents were moved into a new function named `print_network_info()` that takes either a CIDR network address or a network info dict (obtained from `/api/v1/networks/`). 

This change lets us re-use this functionality to print the networks retrieved by `network find`.

The resulting `network.info()` function now looks like this:

https://github.com/unioslo/mreg-cli/blob/f25692a35789da8738f130033252972878e7d3b9/mreg_cli/network.py#L167-L171

and `print_network_info()` has the following signature:

https://github.com/unioslo/mreg-cli/blob/f25692a35789da8738f130033252972878e7d3b9/mreg_cli/network.py#L188


pederhan:
I'll note that I didn't add the documentation for this command to the README, as there is no section on the various `network` commands. Adding documentation for all of those commands is perhaps better suited for a separate pull request.

mreg_cli/network.py
```diff
@@ -203,6 +185,168 @@ def info(args):
...
+def print_network_info(network_info: Union[str, Dict[str, Any]]) -> None:
+    """Prints info about a network given a network address string (CIDR notation),
+    or from a network info dict fetched by `get_network()`.
+
+    If a network address string is passed in, `get_network()` is called to fetch
+    information about the network with the given address.
+    """
+    if isinstance(network_info, str):
+        addr = network_info
+        ip_range = ip_range = get_network_range_from_input(addr)
```
oyvindkolbu, line 197: Double `ip_range`.

mreg_cli/network.py
```diff
@@ -203,6 +185,168 @@ def info(args):
...
 
+def print_network_info(network_info: Union[str, Dict[str, Any]]) -> None:
+    """Prints info about a network given a network address string (CIDR notation),
+    or from a network info dict fetched by `get_network()`.
+
+    If a network address string is passed in, `get_network()` is called to fetch
+    information about the network with the given address.
+    """
+    if isinstance(network_info, str):
+        addr = network_info
```
oyvindkolbu, line 196: Why this `addr` variable?

mreg_cli/network.py
```diff
@@ -203,6 +185,168 @@ def info(args):
...
+
+network.add_command(
+    prog="find",
+    description="Search for networks based on a range of search parameters.",
+    short_desc="Search for networks.",
+    callback=find,
+    flags=[
+        Flag(
+            "-network",
+            description="Network address.",
```
oyvindkolbu, line 292: Looks better without the ending punctation marks in the `desciption` and `short_desc`.

mreg_cli/network.py
```diff
@@ -203,6 +185,168 @@ def info(args):
...
+            type=int,
+        ),
+        Flag(
+            "-silent",
+            description="Do not print meta info (number of networks found, limit reached, etc.).",
+            action="store_true",
+        ),
+    ],
+)
+
```
oyvindkolbu, line 348: `network info` also can find a network given an ip in its range, to match you example `network info 129.199.100.9` .
Maybe an option `-ip` to search for a network with a given IP?
`get_network_range_from_input` not only handles CIDRs or network addresses, but any ip address.

oyvindkolbu (changes requested):

mreg_cli/network.py
pederhan: Since we are shadowing the `network_info` name 2 lines down with the result of `get_network(ip_range)` (thereby changing its type from `str` to `dict`), I added it for ease of debugging down the line should that be needed.

It has no functional impact on the code, so the block can be changed to this if needed:
```py
if isinstance(network_info, str):
    ip_range = get_network_range_from_input(network_info)
    network_info = get_network(ip_range)
```

mreg_cli/network.py
```diff
@@ -236,38 +236,45 @@ def find(args: Namespace):
...
+        if not params:
+            cli_warning("Need at least one search criteria")
+
+        path = f"/api/v1/networks/"
+        networks = get_list(path, params)
+    else:
+        ip_range = get_network_range_from_input(ip_arg)
+        network_info = get_network(ip_range)
+        networks = [network_info]
```
oyvindkolbu, line 267: Classic double negative approach. I'd swap the order and skip the `not`.

mreg_cli/network.py
pederhan: Resolved in e71f4cb.

mreg_cli/network.py
```diff
@@ -276,17 +283,22 @@ def find(args: Namespace):
...
     prog="find",
-    description="Search for networks based on a range of search parameters.",
-    short_desc="Search for networks.",
+    description="Search for networks based on a range of search parameters",
+    short_desc="Search for networks",
     callback=find,
     flags=[
+        Flag(
+            "-ip",
+            description="Exact host address",
```
oyvindkolbu, line 299: Might not be a host. Just an ip.

mreg_cli/network.py
pederhan: [cf93a3b](https://github.com/unioslo/mreg-cli/pull/159/commits/cf93a3b1faba8dcd4089516f3dc174e56e7dd352)

mreg_cli/network.py
pederhan: [af2db9f](https://github.com/unioslo/mreg-cli/pull/159/commits/af2db9ff18cba715ed3f9518c10db8dbc662711f)

oyvindkolbu (approved):

PR #161, opened 2022-11-23
Don't import typing_extensions on runtime 

pederhan:
This pull request removes all imports of `typing_extension` on runtime, and only imports it when [`typing.TYPE_CHECKING`](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) evaluates to True (when running mypy, etc.). This fixes an issue where mreg-cli could not be launched in environments where `typing_extensions` is not available (RHEL7). However, it is still listed as a dependency in setup.py for Python >=3.7.

To not break the existing annotations that make use of `typing_extensions` classes, these annotations are now specified as strings, which is a format mypy and pylance understands. i.e:

```py
# before
def foo(l: List[str]) -> List[str]:
    ...

# after
def foo(l: "List[str]") -> "List[str]":
    ...
```

When Python 3.6 support is dropped, we can instead use `from __future__ import annotations` to prevent type annotations from being evaluated on runtime: https://peps.python.org/pep-0563/

Also, to keep supporting Python 3.6 in the GitHub actions workflow, the workflow OS image has been set to `ubuntu-20.04` instead of `ubuntu-latest`, because the action `actions/setup-python@v2` has dropped support for Python 3.6 in Ubuntu 22.04.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-20.04]
         python-version:
           - '3.6'
           - '3.7'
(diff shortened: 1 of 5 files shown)


oyvindhagberg (approved):
Looks good to me.

PR #164, opened 2023-04-03
Fix workflow annotations and more

oyvindhagberg:
This PR makes some changes to the test workflow:
- Use latest versions of actions
- Add Python 3.10 and 3.11 to the strategy matrix
- Remove unnecessary Postgres version matrix. Testing with different Postgres versions should be done in the server repo instead.
- Add environment vars for the mreg container service

Also, changes to the code:
- A new test for the "host add" command that the mac address is syntactically valid
- Clearer output from policy host_add/host_remove

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,8 @@ jobs:
           - '3.7'
           - '3.8'
           - '3.9'
+          - '3.10'
+          - '3.11'
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -39,12 +41,9 @@ jobs:
   integration:
     name: Integration
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        postgresql-version: [11, 12, 13]
     services:
       postgres:
-        image: postgres:${{ matrix.postgresql-version }}
+        image: postgres:latest
         env:
           POSTGRES_USER: mreg
           POSTGRES_DB: mreg
@@ -61,8 +60,8 @@ jobs:
         options: >-
           --name mreg
         env:
-          GUNICORN_CMD_ARGS: --bind=0.0.0.0
           MREG_DB_HOST: postgres
+          MREG_DB_USER: mreg
           MREG_DB_PASSWORD: postgres
         ports:
           - 8000:8000
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,9 +15,9 @@ jobs:
           - '3.9'
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Cache pip
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: ~/.cache/pip
           key: v1-pip-${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('requirements-*.txt') }}
@@ -26,7 +26,7 @@ jobs:
             v1-pip-${{ runner.os }}
             v1-pip-
       - name: Setup Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install
@@ -68,9 +68,9 @@ jobs:
           - 8000:8000
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Cache pip
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: ~/.cache/pip
           key: v1-pip-${{ runner.os }}-${{ hashFiles('requirements-*.txt') }}
--- a/mreg_cli/policy.py
+++ b/mreg_cli/policy.py
@@ -522,7 +522,7 @@ def host_add(args):
         path = f'/api/v1/hostpolicy/roles/{args.role}/hosts/'
         history.record_post(path, "", data, undoable=False)
         post(path, **data)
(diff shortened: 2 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -39,12 +41,9 @@ jobs:
   integration:
     name: Integration
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        postgresql-version: [11, 12, 13]
     services:
       postgres:
-        image: postgres:${{ matrix.postgresql-version }}
+        image: postgres:latest
```
terjekv, line 46: Will using `:latest` for testing like this lead to us testing an environment that doesn't match what is being run? Wouldn't testing a matrix of versions be safer?

mreg_cli/policy.py
```diff
@@ -522,7 +522,7 @@ def host_add(args):
         path = f'/api/v1/hostpolicy/roles/{args.role}/hosts/'
         history.record_post(path, "", data, undoable=False)
         post(path, **data)
-        cli_info(f"Added {name} to {args.role}", print_msg=True)
+        cli_info(f"Added host {name} to role {args.role}", print_msg=True)
```
terjekv, line 525: Should we quote the variables to make them stand out a bit better? Ala:
`cli_info(f"Added host '{name}' to role '{args.role}'", print_msg=True)`

mreg_cli/policy.py
```diff
@@ -597,7 +597,7 @@ def host_remove(args):
         path = f'/api/v1/hostpolicy/roles/{args.role}/hosts/{name}'
         history.record_delete(path, dict())
         delete(path)
-        cli_info(f"Removed '{name}' from {args.role}", print_msg=True)
+        cli_info(f"Removed host {name} from role {args.role}", print_msg=True)
```
terjekv, line 600: Should we quote the variables to make them stand out a bit better? Ala:
`cli_info(f"Added host '{name}' to role '{args.role}'", print_msg=True)`

testsuite-result.json
```diff
@@ -459,7 +459,7 @@
...
 {"method": "GET", "url": "/api/v1/hosts/foo.example.org", "data": {}, "ok": true, "status": 200, "reason": "OK", "json_data": {"id": 171, "ipaddresses": [], "cnames": [], "mxs": [], "txts": [{"id": 181, "created_at": "2020-12-03T17:04:39.680469+01:00", "updated_at": "2020-12-03T17:04:39.680508+01:00", "txt": "v=spf1 -all", "host": 171}], "ptr_overrides": [], "hinfo": null, "loc": null, "created_at": "2020-12-03T17:04:39.669575+01:00", "updated_at": "2020-12-03T17:04:39.669607+01:00", "name": "foo.example.org", "contact": "", "ttl": null, "comment": "", "zone": 10}}
 {"method": "POST", "url": "/api/v1/hostpolicy/roles/fruit/hosts/", "data": {"name": "foo.example.org"}, "ok": true, "status": 201, "reason": "Created"}
-{"output": "OK: : Added foo.example.org to fruit"}
+{"output": "OK: : Added host foo.example.org to role fruit"}
```
terjekv, line 462: If we quote the variables, this needs to be:
`{"output": "OK: : Added host 'foo.example.org' to role 'fruit'"}`

testsuite-result.json
```diff
@@ -469,7 +469,7 @@
...
 {"method": "GET", "url": "/api/v1/hosts/foo.example.org", "data": {}, "ok": true, "status": 200, "reason": "OK", "json_data": {"id": 171, "ipaddresses": [], "cnames": [], "mxs": [], "txts": [{"id": 181, "created_at": "2020-12-03T17:04:39.680469+01:00", "updated_at": "2020-12-03T17:04:39.680508+01:00", "txt": "v=spf1 -all", "host": 171}], "ptr_overrides": [], "hinfo": null, "loc": null, "created_at": "2020-12-03T17:04:39.669575+01:00", "updated_at": "2020-12-03T17:04:39.669607+01:00", "name": "foo.example.org", "contact": "", "ttl": null, "comment": "", "zone": 10}}
 {"method": "DELETE", "url": "/api/v1/hostpolicy/roles/fruit/hosts/foo.example.org", "data": {}, "ok": true, "status": 204, "reason": "No Content"}
-{"output": "OK: : Removed 'foo.example.org' from fruit"}
+{"output": "OK: : Removed host foo.example.org from role fruit"}
```
terjekv, line 472: If we quote the variables, this needs to be:
`{"output": "OK: : Added host 'foo.example.org' to role 'fruit'"}`

.github/workflows/test.yml
oyvindhagberg: Well. mreg-cli doesn't interface directly with Postgres, mreg does that. So the question becomes, what version of Postgres is mreg using. Both the [test workflow](https://github.com/unioslo/mreg/blob/master/.github/workflows/test.yml#L17) and the [container image workflow](https://github.com/unioslo/mreg/blob/master/.github/workflows/container-image.yml#L36) use `postgres:latest` for testing. What people use in production environments may vary.

The purpose of mreg-cli is to interface with mreg. The way I see it, if mreg doesn't work because of Postgres incompatibilities, thats a problem with mreg not mreg-cli.

terjekv: Fair enough, and besides, if `:latest` actually breaks something in testing, it's the best place to find it early.

mreg_cli/policy.py
oyvindhagberg: Sure, why not. Fixed in [99d4484](https://github.com/unioslo/mreg-cli/pull/164/commits/99d4484d693a2d4ec55cc909a7ac22da650cec23)

Issue #136, opened 2020-11-22
Refactor convert_wildcard_to_*


paalbra:
It probably makes sense to refactor `convert_wildcard_to_filter` and `convert_wildcard_to_regex` after #134 gets resolved.

`convert_wildcard_to_filter` isn't in use. Maybe just remove it?

`convert_wildcard_to_regex` might lead to this pattern:
```python
params = {}
param, value = convert_wildcard_to_regex('name', args.name)
params[param] = value
```
Make input and output a dict probably makes more sense:
```python
params  = convert_wildcard_to_regex({'name': args.name})
```

PR #167, opened 2023-06-07
Clarify help about wildcards, and in some cases automatically use them

oyvindhagberg:
- Improve the help texts that describe when you can use wildcards
- Automatically wrap the parameters in wildcards in some cases
- Add a few commands to the test suite to test the "find"-commands
- Partially takes care of #136.
- Unrelated: Avoid a crash if the token endpoint returns non-json

### DIFF
--- a/mreg_cli/policy.py
+++ b/mreg_cli/policy.py
@@ -385,7 +385,7 @@ def list_atoms(args):
         print("{1:<{0}} {2}".format(padding, key, value))
 
     params = {}
-    param, value = convert_wildcard_to_regex('name', args.name)
+    param, value = convert_wildcard_to_regex('name', args.name, True)
     params[param] = value
     info = get_list("/api/v1/hostpolicy/atoms/", params=params)
     if info:
@@ -402,7 +402,7 @@ policy.add_command(
     callback=list_atoms,
     flags=[
         Flag('name',
-             description='Atom name filter (regexp)',
+             description='Atom name, or part of name. You can use * as a wildcard.',
              metavar='FILTER'),
     ]
 )
@@ -414,7 +414,7 @@ def list_roles(args):
     """
 
     params = {}
-    param, value = convert_wildcard_to_regex('name', args.name)
+    param, value = convert_wildcard_to_regex('name', args.name, True)
     params[param] = value
     info = get_list("/api/v1/hostpolicy/roles/", params=params)
     if not info:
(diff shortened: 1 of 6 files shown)


mreg_cli/policy.py
```diff
@@ -402,7 +402,7 @@ def _print(key, value, padding=20):
     callback=list_atoms,
     flags=[
         Flag('name',
-             description='Atom name filter (regexp)',
+             description='Atom name, or part of name. You can use * as a wildcard.',
```
terjekv, line 405: The text here is a bit unclear. Do I need to use a wildcard to search for part of name? If I need to use a wildcard, the comment should be more like the one on line 310, if not, what purpose does the wildcard serve?

mreg_cli/policy.py
```diff
@@ -445,7 +445,7 @@ def list_roles(args):
     callback=list_roles,
     flags=[
         Flag('name',
-             description='Role name filter (regexp)',
+             description='Role name, or part of name. You can use * as a wildcard.',
```
terjekv, line 448: The text here is a bit unclear. Do I need to use a wildcard to search for part of name? If I need to use a wildcard, the comment should be more like the one on line 310, if not, what purpose does the wildcard serve?

terjekv (changes requested):

mreg_cli/policy.py
terjekv: That works. Although, as the backend is a regex, don't we realistically also support stuff like 'atom_[abc]*'?

terjekv: I honestly find it odd that we rewrite regex at all. From a users perspective I would expect there to be three ways to search for something:

- Exact (equals foo=value, which is equivalent to foo__exact=value)
- Substring (equals foo__contains=value)
- Regex (equals foo_regex=value)

I feel like we are trying to create a query language of our own without a clear syntax to work from, and I wonder what the username actually expects.

oyvindhagberg: Alright so currently we already have the commands `host find`, `network find`, `permission network_list` and `policy list_atoms/list_roles` that support a form of wildcards where a star matches one or more characters. None of the commands support regex, even though the mreg API does.

Feedback from users is that it's unclear where they can use wildcards, and whether it's necessary if supplying only part of a name ( = if the command expexts a substring or exact term).

The aim of this PR was to improve the help texts to clarify that, and make it more consistent across different commands. For the time being, let's see what we can do without rewriting too much.
Many commands expect a substring now, but you can also write things like "foo*bar". In this PR I've tried to make most commands take a substring where it makes sense (instead of the whole term), and change the help texts accordingly.

terjekv: That makes a ton of sense. Thanks!

terjekv (approved):
lgtm

PR #168, opened 2023-06-21
Fix a crash bug when dhcp assoc a host without ips

oyvindhagberg:
This commit fixes a bug where the mreg-cli client would crash if the "dhcp assoc <hostname> <macaddress>" command is used and the given host does not have any ip addresses.

### DIFF
--- a/ci/run_testsuite_and_record.sh
+++ b/ci/run_testsuite_and_record.sh
@@ -14,6 +14,8 @@ trap cleanup EXIT
 docker compose up -d
 
 # create a superuser
+# TODO: Replace this with python manage.py create_mreg_superuser --username test --password test,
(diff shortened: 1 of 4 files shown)


ci/run_testsuite_and_record.sh
```diff
@@ -14,6 +14,8 @@ trap cleanup EXIT
 docker compose up -d
 
 # create a superuser
+# TODO: Replace this with python manage.py create_mreg_superuser --username test --password test,
```
terjekv, line 17: Ah, the joys of inter-repo inter-branch dependencies. :D

terjekv (approved):
Looks good to me. 👍

PR #170, opened 2023-06-27
Better testsuite diffs

oyvindhagberg:


ci/run_testsuite_and_record.sh
```diff
@@ -6,12 +6,13 @@ set -e
...
 # clean up on exit, even if something fails
 function cleanup {
-    docker compose down
+    docker compose --ansi=never down
 }
 trap cleanup EXIT
 
 # start mreg+postgres in containers
-docker compose up -d
+docker compose --ansi=never pull --quiet
```
terjekv, line 14: For some reason, this fails for me with a local image (using `image: mreg/local:latest` in docker-compose.yml). I get `Error response from daemon: pull access denied for mreg/local, repository does not exist or may require 'docker login': denied: requested access to the resource is denied` when using pull and local images... Meh.

For now I simply comment out the pull command when using local images. Ideally we'd also make it easier to do use local images without editing the docker-compose file.

ci/run_testsuite_and_record.sh
```diff
@@ -28,6 +29,8 @@ group.user_set.add(user)
 ' | docker exec -i mreg python /app/manage.py shell
 
 # run the test suite
-echo "test" | mreg-cli -u test -d example.org --url http://127.0.0.1:8000 --source testsuite --record new_output.json >/dev/null
-diff testsuite-result.json new_output.json
+echo "test" | mreg-cli -u test -d example.org --url http://127.0.0.1:8000 --source testsuite --record new_testsuite_log.json >/dev/null
+
+# show a detailed diff
+python diff.py testsuite-result.json new_testsuite_log.json
```
terjekv, line 35: Dear heaven, this is nice. The output is amazing! Thank you!

A few things:

1. I sometimes get the '---' from the diff on the line with the command, like this:

```diff
host sshfp_remove -fingerprint 394875985 bar---
+++
@@ -11,12 +11,17 @@
             "ipaddresses": [
                 {
                     "macaddress": "",
-                    "ipaddress": "10.0.0.14",
+                    "ipaddress": "10.0.0.12",
                     "host": 14
                 },
                 {
-                    "macaddress": "11:22:33:44:55:66",
-                    "ipaddress": "10.0.0.15",
+                    "macaddress": "",
+                    "ipaddress": "2001:db8::11",
+                    "host": 14
+                },
+                {
+                    "macaddress": "11:22:33:44:55:67",
+                    "ipaddress": "2001:db8::12",
                     "host": 14
                 }
             ],
```

2. Should we add a line for each issue regarding which of `+` and `-` that are expected or tested? Or optionally rename them `T` and `E`? Either way, massive improvement and really really helpful.

ci/diff.py
```diff
@@ -0,0 +1,71 @@
...
+    return grouped_objects
+
+
+def main():
+
+	if len(sys.argv) != 3:
+		print("Usage: diff.py <file1> <file2>")
+		sys.exit(1)
+
+	fasit = group_objects(sys.argv[1])
```
terjekv, line 29: Maybe `expected = group_objects(sys.argv[1])`?

ci/diff.py
```diff
@@ -0,0 +1,71 @@
+import difflib,json,sys
```
terjekv, line 1: PEP8 wants imports on separate lines (https://legacy.python.org/dev/peps/pep-0008/#imports).

ci/run_testsuite_and_record.sh
oyvindhagberg: I've modified `run_testsuite_and_record.sh` so it doesn't run `docker compose pull` if run locally, only when run from a github workflow. That should help

ci/run_testsuite_and_record.sh
oyvindhagberg: Thanks! I've made some modifications to `diff.py` now:
- Added a header for each command that should make it more apparent what the pluses and minuses mean
- Fixed the line break problem
- Changed which diff algorithm is used, so the output is more human-readable

oyvindhagberg:
Here's an example of how it looks now, what do you think?
```
========================================================================
Command: group owner_remove mygroup myself        -expected, +tested
========================================================================
  [
      {
          "command": "group owner_remove mygroup myself\n"
      },
      {
          "method": "GET",
          "url": "/api/v1/hostgroups/?name=mygroup",
          "data": {},
          "status": 200,
          "response": {
              "count": 1,
              "next": null,
              "previous": null,
              "results": [
                  {
                      "parent": [],
                      "groups": [],
                      "hosts": [
                          {
                              "name": "testhost1.example.org"
                          }
                      ],
                      "owners": [
                          {
-                             "name": "anotherowner"
?                                      ^^^^^ ^^^^^^
+                             "name": "myself"
?                                      ^^^ ^^
                          },
                          {
-                             "name": "myself"
?                                      ^^^ ^^
+                             "name": "anotherowner"
?                                      ^^^^^ ^^^^^^
                          }
                      ],
                      "name": "mygroup",
                      "description": "This describes the group"
                  }
              ]
          }
      },
      {
          "method": "DELETE",
          "url": "/api/v1/hostgroups/mygroup/owners/myself",
          "data": {},
          "status": 204
      },
      {
          "output": "OK: : Removed 'myself' as owner of 'mygroup'"
      }
  ]
```

ci/run_testsuite_and_record.sh
terjekv: It does indeed help! 👍

terjekv:
This looks very good to me now. Catching dependent errors in an automated way will require a lot more work and I don't think it's worth it right now. :)

terjekv (approved):

PR #171, opened 2023-08-31
Allow for filtering of output.

terjekv:
- Command output may be filtered by `|` followed by whitespace and a regular expression (or a simple string).
  - The regular expression is applied to each line of output, and only lines that match are displayed.
  - The expression may be negated by using `|!` instead of `|`.
  - Substitution is not supported.

Examples:

- `network list_used_addresses 172.16.20.0 | bl23-.*mu2[17]` will output only lines matching the regular expression.
- `network list_used_addresses 172.16.20.0 |! mu2[17]` will output only lines *not* matching the regular expression.

This applies all commands that output to the console, but it does *not* apply to errors, warnings, or informational messages. This allows filters to be used without concern for missing important information.


mreg_cli/main.py
```diff
@@ -164,17 +171,33 @@ def main():
...
+                filter_re = None
+                negate = False
+
+                if "|" in line:
+                    command, filter_str = line.split("|", 1)
+                    filter_str = filter_str.strip()
+
+                    if filter_str.startswith("!"):
+                        negate = True
+                        filter_str = filter_str[1:].strip()
```
oyvindkolbu, line 187: This strip is superflous as it has already been stripped. Just need the `[1:]`.

oyvindkolbu:
You are mixing functial and formatting changes. Makes review very hard.

terjekv:
Yeah. Sadly my editor setup is strict with PEPs. I'll try to review the settings for the project for every copy of the repo I work on. In general though I'm tempted to do a pure formatting PR towards black and isort, and adding them to a lint job.

terjekv:
Superseded by a much better implementation in #174.

PR #172, opened 2023-09-02
Linting and formatting, semi-lax.

terjekv:
### Initial linting and formatting adherence

- `tox` now runs `black` and `ruff` on the project (the environment is called `lint`).
- tox runs `ruff` with --fix to resolve fixable issues.
- CI integration for tox is added to the test job.

### Code changes

To adhere to even semi-lax linting, there were changes required in the code. These are:

- Migrated use of python builtin identifiers (`type`,  `id`, and `print`) to more appropriate names.
- Some cases of long string literals were using implicit concatenation, which is now explicit (by adding a set of parentheses)
- Note that unused imports without `# noqa: F401` following the import will be removed by ruff. 
- Exception handling demanded work as there were unspecified exception catching (`except Exception` rewritten to explicit exceptions) and cases where exceptions within exceptions did not specify if one was to include the previous exception or not (use `from` either from the previous exception or `None` explicitly).

### Outstanding changes with current settings

- `print` in history.py (shadowing builtin) was not fixed as it needs a bit more debugging to figure out exactly how it is used.

### Notes

There is still a lot to do with regards to linting, formatting, and typing. This PR does not seek to resolve these issues, but attempts to create a baseline of requirements for the codebase to adhere to.


oyvindhagberg (approved):
This was a lot to read through, with all the whitespace and quotation mark changes. But it looks good to me ™️

PR #174, opened 2023-11-12
feat: Allow for filtering of output

terjekv:
- Command output may be filtered by `|` followed by whitespace and a regular expression (or a simple string).
- The regular expression is applied to each line of output, and only lines that match are displayed.
- The expression may be negated by using `|!` instead of `|`.
- Substitution is not supported.

Examples:

- `network list_used_addresses 172.16.20.0 | bl23-.*mu2[17]` will output only lines matching the regular expression.
- `network list_used_addresses 172.16.20.0 |! mu2[17]` will output only lines *not* matching the regular expression.

This applies all commands that output to the console, but it does *not* apply to errors, warnings, or informational messages. This allows filters to be used without concern for missing important information.

This is a cleaner Singleton-based implementation of #171.

### DIFF
--- a/mreg_cli/outputmanager.py
+++ b/mreg_cli/outputmanager.py
@@ -1,0 +1,128 @@
+"""This module contains the output management for the mreg_cli package.
+
+This is a singleton class that manages the output for the CLI. It stores the
+output lines and formats them for display. It also manages the filter for the
+command.
+"""
+
+import re
+from typing import Any, List, Tuple
+
+
+class OutputManager:
+    """Manages and formats output for display.
+
+    This is a singleton class to retain context between calls. It is initiated
+    and cleared in the main function of the CLI, all other calls should only
+    add lines to the output via add_line(), add_formatted_line(), or
+    add_formatted_line_with_source(),
+    """
+
+    _instance = None
+
+    def __new__(cls):
+        if cls._instance is None:
+            cls._instance = super(OutputManager, cls).__new__(cls)
+            cls._instance._lines = []
+        return cls._instance
+
+    def clear(self) -> None:
+        """Clears the object."""
+        self._lines = []
+        self._filter_re = None
+        self._negate = False
+        self._command = None
+
+    def has_output(self) -> bool:
+        """Returns True if there is output to display."""
+        return len(self._lines) > 0
+
+    def from_command(self, command: str) -> str:
+        """Adds the command that generated the output.
+
+        :param command: The command to add.
+
+        :return: The cleaned command, devoid of filters and other noise.
+        """
+        self.command = command
+        self._command, self._filter_re, self._negate = self.get_filter()
+        return self._command
+
+    def add_line(self, line: str) -> None:
+        """Adds a line to the output.
+
+        :param line: The line to add.
+        """
+        self._lines.append(line)
+
+    def add_formatted_line(self, key: str, value: str, padding: int = 14) -> None:
+        """Formats and adds a key-value pair as a line.
+
+        :param key: The key or label.
+        :param value: The value.
+        :param padding: The padding for the key.
+        """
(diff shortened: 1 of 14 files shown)


mreg_cli/outputmanager.py
```diff
@@ -0,0 +1,128 @@
...
+            cls._instance = super(OutputManager, cls).__new__(cls)
+            cls._instance._lines = []
+        return cls._instance
+
+    def clear(self) -> None:
+        """Clears the object."""
+        self._lines = []
+        self._filter_re = None
+        self._negate = False
+        self._command = None
```
pederhan, line 34: Calling `get_filter()` by itself or before `from_command()` will raise an `AttributeError` due to the attribute `command` not being set. Consider setting the attribute in `clear()`:

```suggestion
        self._command = None
        self.command = None
```

mreg_cli/outputmanager.py
```diff
@@ -0,0 +1,128 @@
...
+        self.add_line(formatted_line)
+
+    # We want to use re.Pattern as the type here, but python 3.6 and older re-modules
+    # don't have that type. So we use Any instead.
+    def get_filter(self) -> Tuple[str, Any, bool]:
+        """Returns the filter for the output.
+
+        :return: The filter and whether it is a negated filter.
+        """
+        command = self.command
```
pederhan, line 88: What if the command was passed in as an argument instead, so that the method isn't reliant on `from_command()` setting the `self.command` attribute? Easier to test, etc.

pederhan (commented):
LGTM (with some caveats)

I have some minor nitpicks that pertain to the internal state of the `OutputManager` class, and how it can be changed to improve its testability and make it more robust. 

I can see that how it's used now is totally fine, but I am generally not a huge fan of an object's internal state dictating whether or not it's safe to call a method - especially when you might want to test said methods in isolation.

mreg_cli/outputmanager.py
terjekv: I have no recollection of why it ended up this way. It is much cleaner to pass things explicitly. I'll retain the self.command for now though, but assign it in get_filter instead. Thanks!

mreg_cli/outputmanager.py
terjekv: Indeed! Fixed.

mreg_cli/outputmanager.py
```diff
@@ -0,0 +1,128 @@
...
+    add lines to the output via add_line(), add_formatted_line(), or
+    add_formatted_line_with_source(),
+    """
+
+    _instance = None
+
+    def __new__(cls):
+        if cls._instance is None:
+            cls._instance = super(OutputManager, cls).__new__(cls)
+            cls._instance._lines = []
```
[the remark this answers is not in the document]
terjekv: Yep, good point! Fixed.

mreg_cli/outputmanager.py
```diff
@@ -0,0 +1,128 @@
...
+    and cleared in the main function of the CLI, all other calls should only
+    add lines to the output via add_line(), add_formatted_line(), or
+    add_formatted_line_with_source(),
+    """
+
+    _instance = None
+
+    def __new__(cls):
+        if cls._instance is None:
+            cls._instance = super(OutputManager, cls).__new__(cls)
```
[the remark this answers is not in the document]
terjekv: I'm picking up bad habits from old code. Thanks!

pederhan (approved):
LGTM!

oyvindhagberg:
Good work! @terjekv if you can just rebase this on top of master, I think we can merge it. Optionally squash

terjekv:
Rebased on top of master. :)

terjekv:
Btw, I didn't find any nice way to document this via `-h` once in the cli (which also lies about exiting the application). I think we may want to write a proper help command for generalised cli help.

PR #180, opened 2023-11-29
fix: Handle unquoted pipes.

terjekv:
- Prevents the CLI from dying when confronted with commands like `permission network_add 123.123.123.0/27 some-group-name ^[a-z0-9\-]+(\.(aa|bbb|ccc|sikkerhet?\.uio\.no$`
  - To allow for reuse of exception formatting, moves the formatting to the exceptions themselves rather than leaving the issue for the user.


pederhan (approved):
LGTM

PR #182, opened 2023-12-01
feat: Merge Recorder and OutputManager.

### LINKED ISSUE #179: Test the new functionality introduced in #174
- Test the new functionality introduced in #174
  - Add commands to the testsuite to test `permission network_add` with a regex that contains the characters | and |! , as a [regression test for this commit](https://github.com/unioslo/mreg-cli/commit/f11035156016859cda3b319b5f9f42bd02b3f5e9)
  - Update `testsuite-result.json` (use `run_testsuite_and_record.sh` and look at `new_testsuite_log.json`, comparing the files)
- The `source` command / cmdline argument must use OutputManager
  - There should be no difference in how output is handled when using the `source` command/argument and by typing commands interactively
- Optionally, merge Recorder and OutputManager

### LINKED ISSUE #181: Recorder does not record tables.
Table output is not recored properly. This may have been introduced with the use of OutputManager, but seems to have been around for even longer than that. This leads to recording output like this:

```json
{
    "command": "permission network_list\n"
  },
  {
    "method": "GET",
    "url": "/api/v1/permissions/netgroupregex/?ordering=range%2Cgroup",
    "data": {},
    "status": 200,
    "response": {
      "count": 1,
      "next": null,
      "previous": null,
      "results": [
        {
          "group": "mygroup",
          "range": "192.168.0.0/16",
          "regex": ".*",
          "labels": [
            2
          ]
        }
      ]
    }
  },
  {
    "method": "GET",
    "url": "/api/v1/labels/",
    "data": {},
    "status": 200,
    "response": {
      "count": 1,
      "next": null,
      "previous": null,
      "results": [
        {
          "name": "postit",
          "description": "A label again"
        }
      ]
    }
  },
  {
    "command": "label info postit\n"
  },
  {
    "method": "GET",
    "url": "/api/v1/labels/name/postit",
    "data": {},
    "status": 200,
    "response": {
      "name": "postit",
      "description": "A label again"
    }
  },
  {
    "method": "GET",
    "url": "/api/v1/hostpolicy/roles/?labels__name=postit",
    "data": {},
    "status": 200,
    "response": {
      "count": 0,
      "next": null,
      "previous": null,
      "results": []
    }
  },
```

terjekv:
This PR seeks to merge Recorder and OutputManager into a unified interface for processing commands and producing output, either to file or console. This ensures that:

  - Filtering (`|`or `|!`) works from commands called via `source`
  - Output to files is filtered as expected.

## Structural changes

  - Removes `recorder.py` and its class.
  - Incorporates recording into OutputManager.
  - The OutputManager handles recording (mostly) behind the scenes if we're in recording mode. Outside of log handling we don't have to call recording tools explicitly as its done as part of normal output processing.
  - Incorporates table formatting into OutputManager where it belongs.
  - Moves command processing to a shared separate function called from both `source` and the interactive loop.

## CLI changes

  - Adds commands to start and stop recording interactively.
  - Updates the prompt to show active recording if applicable.

## (Previously) Failing tests!

This PR did not pass the old tests as it fixes #181. After these changes were thourughly reviewed and verified, the new testsuite results were updated in https://github.com/unioslo/mreg-cli/pull/182/commits/9fe78e0373b8b0d9af8cf56e5cb90eccf4e34001.

## Other thoughts

  - We should revise the JSON format currently employed. Right now there is only sequential implied logic between commands and their output. A user of the output has to assume that for every `command` element in the log, there will be any number of API calls, and then eventually, hopefully, an `output` element. After that, the cycle starts over again. This is not very parsable. At all. There is no structural relationship between the command, the API calls, and the output. 
  - We may want to offer a simpler output format to flat files. People who don't love `jq` will probably prefer that, and they can use filtering to massage the results inline.
  - We probably want to add an append mode to the recording file.
  - Warnings and Errors have a weird double colon (`WARNING: : host not found: 'foo'`), I'm unsure if there is a good reason for this choice(?).

Fixes #179, fixes #181.


pederhan:
> We should revise the JSON format currently employed. Right now there is only sequential implied logic between commands and their output.

Agreed. There should be some coupling between each command and its requests and output.

pederhan (approved):
Works as expected without any issues for me.

LGTM

pederhan:
> Should we look at that as a part of this patch, or as a separate PR?

Up to you. I think this PR is good as is, and doesn't _need_ to address that problem to be considered "complete".

PR #187, opened 2023-12-07
Linting updates.

terjekv:
## Changes

Ruff rules added:

  - Add "D": pydocstyle.
  - Add "F": Pyflakes

Also fix all hard-failing type definitions.

## Outstanding

Pylint-refactor (PLR) and pylint-warning (PLW) should be done next, but they require code refactoring, and this PR tries hard to avoid anything more than cosmetics.

### DIFF
--- a/mreg_cli/types.py
+++ b/mreg_cli/types.py
@@ -3,7 +3,7 @@ import ipaddress
 import sys
 from typing import TYPE_CHECKING, Union
 
-# Horrible heck to support Literal on Python 3.6
+# Horrible hack to support Literal when possible.
 if sys.version_info >= (3, 8):
     from typing import Literal
 
--- a/mreg_cli/types.py
+++ b/mreg_cli/types.py
@@ -1,4 +1,10 @@
-from typing import TYPE_CHECKING
+"""Typing definitions for mreg_cli."""
+import ipaddress
+from typing import TYPE_CHECKING, Literal, Union
+
+IP_network = Union[ipaddress.IPv4Network, ipaddress.IPv6Network]
+IP_Version = Literal[4, 6]
+
 
 if TYPE_CHECKING:
     from typing import Any
@@ -10,15 +16,19 @@ if TYPE_CHECKING:
 
         @property
         def ok(self) -> bool:
+            """Return True if the response was successful."""
             ...
 
         @property
         def status_code(self) -> int:
+            """Return the HTTP status code."""
             ...
 
         @property
         def reason(self) -> str:
+            """Return the HTTP status reason."""
             ...
 
         def json(self, **kwargs: Any) -> Any:
+            """Return the response body as JSON."""
             ...
--- a/mreg_cli/types.py
+++ b/mreg_cli/types.py
@@ -1,9 +1,12 @@
 """Typing definitions for mreg_cli."""
 import ipaddress
-from typing import TYPE_CHECKING, Literal, Union
+from typing import TYPE_CHECKING, Union
 
 IP_network = Union[ipaddress.IPv4Network, ipaddress.IPv6Network]
-IP_Version = Literal[4, 6]
+# IP_Version = Literal[4, 6]
+# IP_Version can be either 4 or 6, but Literal is not supported in older versions of Python.
+# We could use typing_extensions.Literal, but that would require an extra dependency.
+IP_Version = int
 
 
 if TYPE_CHECKING:
--- a/mreg_cli/types.py
+++ b/mreg_cli/types.py
@@ -4,8 +4,8 @@ import sys
 from typing import TYPE_CHECKING, Union
 
 # Horrible heck to support Literal on Python 3.6
-if sys.version_info >= (3, 7):
-    from typing_extensions import Literal
+if sys.version_info >= (3, 8):
+    from typing import Literal
 
     IP_Version = Literal[4, 6]
 else:
--- a/mreg_cli/types.py
+++ b/mreg_cli/types.py
@@ -1,12 +1,19 @@
 """Typing definitions for mreg_cli."""
 import ipaddress
+import sys
 from typing import TYPE_CHECKING, Union
 
+# Horrible heck to support Literal on Python 3.6
+if sys.version_info >= (3, 7):
+    from typing_extensions import Literal
+
+    IP_Version = Literal[4, 6]
+else:
+    from typing import int
+
+    IP_Version = int
+
 IP_network = Union[ipaddress.IPv4Network, ipaddress.IPv6Network]
-# IP_Version = Literal[4, 6]
-# IP_Version can be either 4 or 6, but Literal is not supported in older versions of Python.
-# We could use typing_extensions.Literal, but that would require an extra dependency.
-IP_Version = int
 
 
 if TYPE_CHECKING:
--- a/mreg_cli/types.py
+++ b/mreg_cli/types.py
@@ -1,4 +1,5 @@
 """Typing definitions for mreg_cli."""
+
 import ipaddress
 import sys
 from typing import TYPE_CHECKING, Union
--- a/mreg_cli/types.py
+++ b/mreg_cli/types.py
@@ -9,8 +9,6 @@ if sys.version_info >= (3, 8):
 
     IP_Version = Literal[4, 6]
 else:
-    from typing import int
-
     IP_Version = int
 
 IP_network = Union[ipaddress.IPv4Network, ipaddress.IPv6Network]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,10 @@ jobs:
       - name: Install mreg-cli
         run: |
           pip install -r requirements.txt
+          pip install -r requirements-dev.txt
           pip install -e .
+      - name: Check linting
+        run: tox -e lint
       - name: Test and compare api calls
         run: ci/run_testsuite_and_record.sh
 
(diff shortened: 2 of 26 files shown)


mreg_cli/types.py
```diff
@@ -1,4 +1,13 @@
-from typing import TYPE_CHECKING
+"""Typing definitions for mreg_cli."""
+import ipaddress
+from typing import TYPE_CHECKING, Union
+
+IP_network = Union[ipaddress.IPv4Network, ipaddress.IPv6Network]
+# IP_Version = Literal[4, 6]
+# IP_Version can be either 4 or 6, but Literal is not supported in older versions of Python.
+# We could use typing_extensions.Literal, but that would require an extra dependency.
```
pederhan, line 8: We already have `typing_extensions` as a dependency for Python >3.6 (blame me), however, we excluded it for Python3.6 because it was not available in EPEL 7. Importing from `typing_extensions` on runtime would cause the program to crash when installed via the RPM package on RHEL 7.

Now, RHEL 7 is not very widely used, so we could decide to drop support for it, but a less breaking change would be to do something like this:

```py
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from typing_extensions import Literal

    IP_Version = Literal[4, 6]


# in some other module:
if TYPE_CHECKING:
    from .types import IP_Version


def takes_version(version: "IP_Version") -> None:
    pass
```

However, that means it's necessary to always import `IP_Version` in `if TYPE_CHECKING: ...` blocks and use it as a string literal annotation, which is not super elegant either.

An alternative that doesn't require importing in `if TYPE_CHECKING: ...` is something like this:

```py
try:
    from typing_extensions import Literal

    IP_Version = Literal[4, 6]
except ImportError:
    IP_Version = int  # type: ignore
```

Which isn't great either.

In general, typing in Python <3.7 is pretty painful due to the lack of `from __future__ import annotations`, and doubly so when we can't even use `typing_extensions`.

pederhan (approved):

mreg_cli/types.py
terjekv: I went for the latter option. Ish.

```python
# Horrible hack to support Literal when possible.
if sys.version_info >= (3, 8):
    from typing import Literal

    IP_Version = Literal[4, 6]
else:
    IP_Version = int
```

pederhan: Oh yeah, that works better. I assumed mypy would complain about that, but it didn't. 

I tried this, and it didn't work:

```py
try:
    from typing import Literal

    IP_Version = Literal[4, 6]
except ImportError:
    IP_Version = int
```

`error: Cannot assign multiple types to name "IP_Version" without an explicit "Type[...]" annotation`

The explicit conditional in your solution seems to create 2 different scopes, which allows for 2 different type assignments.

terjekv: Yeah, both branches of `try` seems to be evaluated, which is... I don't know. Something.

pederhan (approved):
LGTM, can be merged when you are ready.

mreg_cli/types.py
pederhan: I've been thinking about this problem since we discussed it...

Using a `TypeAlias` as a string annotation for a string containing the `Literal` definition, wherein both `Literal` and `TypeAlias` are imported in an `if TYPE_CHECKING: ...` block seems to be the most compatible solution and has no runtime impact.

```py
# mypy_type_alias.py

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from typing_extensions import Literal  # noqa: F401
    from typing_extensions import TypeAlias

IP_Version: "TypeAlias" = "Literal[4, 6]"


def takes_version(version: "IP_Version") -> None:
    pass
```

```
% for i in {6..12}; do echo -n "3.$i: "; mypy --python-version 3.$i mypy_type_alias.py; done
3.6: Success: no issues found in 1 source file
3.7: Success: no issues found in 1 source file
3.8: Success: no issues found in 1 source file
3.9: Success: no issues found in 1 source file
3.10: Success: no issues found in 1 source file
3.11: Success: no issues found in 1 source file
3.12: Success: no issues found in 1 source file
```

One could also import from `typing` (since we never actually attempt to evaluate the annotations on runtime), but in that case mypy will complain about previous Python versions, even though it's OK in practice:

```py
if TYPE_CHECKING:
    from typing import Literal  # noqa: F401
    from typing import TypeAlias
```

```
% for i in {6..12}; do echo -n "3.$i: "; mypy --python-version 3.$i mypy_type_alias.py; done
3.6: mypy_type_alias.py:4: error: Module "typing" has no attribute "Literal"
mypy_type_alias.py:5: error: Module "typing" has no attribute "TypeAlias"
Found 2 errors in 1 file (checked 1 source file)
3.7: mypy_type_alias.py:4: error: Module "typing" has no attribute "Literal"
mypy_type_alias.py:5: error: Module "typing" has no attribute "TypeAlias"
Found 2 errors in 1 file (checked 1 source file)
3.8: mypy_type_alias.py:5: error: Module "typing" has no attribute "TypeAlias"
Found 1 error in 1 file (checked 1 source file)
3.9: mypy_type_alias.py:5: error: Module "typing" has no attribute "TypeAlias"
Found 1 error in 1 file (checked 1 source file)
3.10: Success: no issues found in 1 source file
3.11: Success: no issues found in 1 source file
3.12: Success: no issues found in 1 source file
```

```
python3.6 mypy_type_alias.py; echo $?
0
```


But to be honest, the cleanest solution is to just drop RHEL 7 support (there is no `python3-typing-extensions` package in EPEL 7). That way we can require `typing_extensions` for every Python version, instead of exempting Python 3.6 from the requirement.

oyvindkolbu: Just drop RHEL 7 support. It is soon on extended life support anyhow and should be avoided.

terjekv: String annotations always makes me ick a bit tbh, it's like using strings for model names in Django to prevent circular dependencies. You don't notice the error before it blows in a bad way, and one becomes reliant on static linters being _very_ smart to give good error messages. With that being said, I don't really understand what python3.6 will do with this example:

```python
# mypy_type_alias.py

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from typing_extensions import Literal  # noqa: F401
    from typing_extensions import TypeAlias

IP_Version: "TypeAlias" = "Literal[4, 6]"


def takes_version(version: "IP_Version") -> None:
    pass
```

As in, what type does python3.6 think IP_Version actually is? It passes mypy, sure, but... Of course, it doesn't *really* matter as typing is an IDE / linter helper in python anyway, and none of us (I dearly hope) are developing with python 3.6 anyway so...

I'm still a bit sad from reading https://peps.python.org/pep-0484/  and seeing:

> While these annotations are available at runtime through the usual `__annotations__` attribute, _no type checking happens at runtime_.

terjekv: > Just drop RHEL 7 support. It is soon on extended life support anyhow and should be avoided.

Well, yes, but...

```
terjekv@...-rhel8-prod01:~$ python3 --version
Python 3.6.8
```

pederhan: I wrote a long reply, and misclicked "resolve conversation"... TL;DR: 

> As in, what type does python3.6 think IP_Version actually is?

```
> /Users/pederhan/repos/misc/mypy_type_alias.py(16)<module>()->None
-> pdb.set_trace()
(Pdb) takes_version.__annotations__
{'version': 'IP_Version', 'return': None}
(Pdb) IP_Version
'Literal[4, 6]'
(Pdb) type(IP_Version)
<class 'str'>
(Pdb) 
```

pederhan: > > Just drop RHEL 7 support. It is soon on extended life support anyhow and should be avoided.
> 
> Well, yes, but...
> 
> ```
> terjekv@...-rhel8-prod01:~$ python3 --version
> Python 3.6.8
> ```
> 
> At least we get working type extensions...

Yeah, so we are still unable to use `from __future__ import annotations` to resolve stuff like `re.Pattern` without string annotations, but the backports of `Literal`, `TypeAlias`, etc. are hugely helpful.

terjekv: Makes sense. I always wanted an IP version "xyzzy". :)

pederhan: > Makes sense. I always wanted an IP version "xyzzy". :)

To be clear, it's not a string and mypy _will_ complain even when targetting 3.6, since the type checking machinery is independent of the actual Python runtime:

```py
takes_version("xyzzy")
takes_version(4)
takes_version(6)

IP_Version = str
IP_Version = "foo"
```

```
% mypy --python-version 3.6 mypy_type_alias.py
mypy_type_alias.py:14: error: Argument 1 to "takes_version" has incompatible type "Literal['xyzzy']"; expected "Literal[4, 6]"
mypy_type_alias.py:18: error: Cannot assign multiple types to name "IP_Version" without an explicit "Type[...]" annotation
mypy_type_alias.py:19: error: Cannot assign multiple types to name "IP_Version" without an explicit "Type[...]" annotation
Found 3 errors in 1 file (checked 1 source file)
```

pederhan: There is a Ruff setting for this that automates the process as long as `from __future__ import annotations` is not imported (which it will never be for us):

https://docs.astral.sh/ruff/settings/#flake8-type-checking-quote-annotations

terjekv: Oh! Sweet. Thanks!

oyvindkolbu: > > Just drop RHEL 7 support. It is soon on extended life support anyhow and should be avoided.
> 
> Well, yes, but...
> 
> ```
> terjekv@...-rhel8-prod01:~$ python3 --version
> Python 3.6.8
> ```

Though rhel 8 provides python 3.9 as well. Just add it as a requirement.

PR #188, opened 2023-12-08
Prevent dynamic dates from being included in test suite.

terjekv:
- Info for policy and roles returns created date.
  - The simple solution is to filter those results out of the suite.

This will again break the current tests...

### DIFF
--- a/ci/testsuite-result.json
+++ b/ci/testsuite-result.json
@@ -8504,7 +8504,7 @@
     "output": "Name:          fruit\nDescription:   5 a day\nAtom members:\n               apple\n               orange\nLabels:\n               None"
   },
   {
-    "command": "policy list_members fruit\n"
+    "command": "policy list_members fruit |! Created"
   },
   {
     "method": "GET",
--- a/ci/testsuite-result.json
+++ b/ci/testsuite-result.json
@@ -98,7 +98,7 @@
     }
   },
   {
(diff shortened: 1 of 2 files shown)


ci/testsuite-result.json
```diff
@@ -98,7 +98,7 @@
     }
   },
   {
-    "output": "Zone:               example.org\n        Nameservers:        hostname            TTL\n                            ns1.example.org     <not set>\nPrimary ns:         ns1.example.org\nEmail:              hostmaster@example.org\nSerialnumber:       3231206000\nRefresh:            10800\nRetry:              3600\nExpire:             1814400\nSOA TTL:            43200\nDefault TTL:        43200"
+    "output": "Zone:               example.org\n        Nameservers:        hostname            TTL\n                            ns1.example.org     <not set>\nPrimary ns:         ns1.example.org\nEmail:              hostmaster@example.org\nSerialnumber:       3231208000\nRefresh:            10800\nRetry:              3600\nExpire:             1814400\nSOA TTL:            43200\nDefault TTL:        43200"
```
terjekv, line 101: What I don't understand is why the Serialnumber for the Zone changed. :(

pederhan: Weird, but nothing in this PR should have changed that, and the tests pass in CI...

pederhan (approved):
LGTM

ponas:
https://github.com/spulec/freezegun could help here

terjekv:
Looking at something like freezegun in the longer term is probably not a bad idea. We're not doing any comparisons between times and dates on the client and server anyway, so it should be safe. Yes, I have had the misfortune of debugging why mocking dates didn't work, only to realise I was checking against a timestamp generated by the server, who had zero understanding of my desire to mock dates. :)

PR #189, opened 2023-12-11
Remove zone serialnumber from test output.

terjekv:
- When the serial number for zones is created (https://github.com/unioslo/mreg/blob/master/mreg/utils.py#L78) it is created with as a "10-digit serialnumber in 3YYMMDDXXX format".
  - It's hard to mock this as it comes from the server and it's not like we're in a pytest environment with the ci runs here.
  - We would need to say something like "run this entire cli session as if the date and time was as follows", and outside of modifying the cli itself (it's an insane thing to offer as an option), we'd have to modify the runtime environment (ie, the container), but that would break SSL and its friends...
  - So, in short, we remove Serialnumber from the test output...


pederhan (approved):
LGTM

PR #191, opened 2023-12-17
Major, _major_, refactoring.

terjekv:
(Note: This PR is *huge*. Merging is unlikely, but it may be a useful target for development.)

## Goals met

  - Avoids all magic imports (commands).
  - Avoids globals (config).
  - Avoids importing the cli itself all over the place.
  - Avoids hacking around circular imports via inline imports (commands again).
  - Massively cleans up utility functions, making an attempt at standardising their naming.
  - Cleans up the code around commands and their declaration / callback mishmash.

## Big picture changes

  - Split util.py into modules under mreg_cli.utilities.
  - Move command modules in under mreg_cli.commands.
  - Split host commands into smaller bits under mreg_cli.commands.host_submodules
  - Use decorators for commands instead of following callbacks with command additions.
  - Create a proper configuration module (config.py).
  - Set up a help command.
  - Update README.md to match reality.
  - Remove the broken history implementation. This is not to be confused with the server side object history commands (`host history foo`). This is the `history` command on at the root which shows what happened this session and if you can undo (or redo) the operation. The server side lookups are useful. The local what-have-I-done-in-this-session can do with a much much better implementation if anyone wants it.
  - Adds a Help formatter for the cli output, this ensures that help output avoids the line breaks due to long command names. 

## Decorator implementation

Command modules create a command class that inherits from `mreg_cli.commands.base.BaseCommand`. This class is initialized with a `mreg_cli.commands.registry.CommandRegistry` instance (ie, the decorator) to produce a local decorator for its scope. As an example:

```python
[...]
command_registry = CommandRegistry()


class DHCPCommands(BaseCommand):
    """DHCP commands for the CLI."""

    def __init__(self, cli: Any) -> None:
        """Initialize the DHCP commands."""
        super().__init__(cli, command_registry, "dhcp", "Manage DHCP associations", "Manage DHCP")


@command_registry.register_command(
    prog="assoc",
    description=(
        "Associate MAC address with a host. If the host has multiple A/AAAA "
        "records an IP must be given instead of name."
    ),
    short_desc="Add MAC address to host.",
    flags=[
        Flag("name", "Name or IP of target host.", metavar="NAME/IP"),
        Flag("mac", "Mac address.", metavar="MACADDRESS"),
        Flag("-force", action="store_true", description="Enable force."),
    ],
)
def assoc(args: argparse.Namespace) -> None:
    """Associate MAC address with host.

    If the host has multiple A/AAAA records an IP must be given instead of a name.

    :param args: argparse.Namespace (name, mac, force)
    """
    ip = get_unique_ip_by_name_or_ip(args.name)
    new_mac = assoc_mac_to_ip(args.mac, ip, force=args.force)

    if new_mac is not None:
        cli_info(
            "associated mac address {} with ip {}".format(new_mac, ip["ipaddress"]),
            print_msg=True,
        )
```

## Configuration module

Supports input from a configuration file, environment, and comand line and prioritises the keys in that order.

## Help command

A help command is now available. At the root, it does the same as `-h`, but it also allows for subcommands. Currently supported subcommands:

  - `filtering` provides help about the filtering operators, with examples.
  - `configuration` shows the configuration options currently being employed.

## Goals still outstanding

  - Further cleanup of utility functions. There are still some scattered around here and there.
  - Improve test suite.
  - Improve the recording system, extend to use as a debug / history that works.
    - Record a sliding window of all calls and commands.
    - Allowing the user to toggle debugging output to the terminal or to a file.
      - Debugging output can be verbosity up to and including API calls and their entire output.
    - When recording, we need to do a proper append cycle to the log file to prevent potential infinite memory usage.
  - Clean up a lot of the commands, especially the RR commands. There is a lot of copy-paste that we can sort.
  - Clean up the API calls. Apart from using `get_list` to get individual items, we've got a ton of literal strings as URLs scattered around everywhere. Only the API code should know anything about endpoints.
  - Clean up the API code itself. Fiddling with internal variables to fetch and/or manipulate data is horrific.


mreg_cli/commands/registry.py
```diff
@@ -0,0 +1,52 @@
...
+        """Initialize the command registry."""
+        self._commands: List[Command] = []
+
+    def register_command(
+        self,
+        prog: str,
+        description: str,
+        short_desc: str,
+        flags: Optional[List[Flag]] = None,
+    ) -> Callable[[argparse.Namespace], None]:
```
pederhan, line 31: Currently, mypy gives us the following error:

```
mreg_cli/commands/zone.py:127: error: Argument 1 has incompatible type "Callable[[Namespace], None]"; expected "Namespace"
```

The current annotation claims this method returns a command function, which is incorrect. The `register_command()` method actually returns a decorator function that takes a command function which THEN returns said command function. 


It might be useful to add an alias for command functions to `types.py`:

```py
# types.py
CommandFunc = Callable[[argparse.Namespace], None]
```

Which would allow us to do this:

```suggestion
    ) -> Callable[[CommandFunc], CommandFunc]:
```

as opposed to the more verbose:

```suggestion
    ) -> Callable[[Callable[[argparse.Namespace], None]], Callable[[argparse.Namespace], None]]:
```

terjekv: Yes, thank you! I'm disappointed that pylance didn't catch that return issue. Bleeding obvious when you look at it though. Fixed in https://github.com/terjekv/mreg-cli/commit/38032ccb70ef7e21a03ee1d2898d4abd7ff1a792 (but there are lots of other mypy issues as well, mostly inherited.)

mreg_cli/utilities/api.py
```diff
@@ -0,0 +1,270 @@
...
+
+    if use_json:
+        result = getattr(session, operation_type)(url, json=params, timeout=HTTP_TIMEOUT)
+    else:
+        result = getattr(session, operation_type)(
+            url, params=params, data=data, timeout=HTTP_TIMEOUT
+        )
+    result = cast(requests.Response, result)  # convince mypy that result is a Response
+
+    OutputManager().recording_request(operation_type, url, params, data, result)
```
pederhan, line 187: The `params` parameter of `OutputManager.recording_request` takes a `str` arg, yet we pass it a `dict` here?

mreg_cli/utilities/api.py
```diff
@@ -0,0 +1,270 @@
...
+mreg_auth_token_file = os.path.join(str(os.getenv("HOME")), ".mreg-cli_auth_token")
+
+logger = logging.getLogger(__name__)
+
+HTTP_TIMEOUT = 20
+
+config = MregCliConfig()
+
+
+mregurl = config.get("url")
```
pederhan, line 40: This value _can't_ be `None`. All the calls to `urljoin` will fail we pass it `None`, as I understand it.

terjekv: Fixed in https://github.com/unioslo/mreg-cli/pull/191/commits/67956443bf3c4ae1ade965238bea5298ee700c26

mreg_cli/utilities/api.py
terjekv: This seems to have been resolved at some point?

pederhan: Must have been a result of the type annotation fixes in my PR.

pederhan:
```
% mypy mreg_cli           
Success: no issues found in 37 source files
```
Perfect. 

----

I have created a new branch `legacy`, so we can merge this into master while also supporting bug fixes for the current version.

pederhan (approved):

PR #197, opened 2024-01-11
bug: Excluded addresses not counted correctly.

### LINKED ISSUE #196: Excluded addresses not counted correctly
Running "network info <network>" on a network with excluded addresses miscalculates the number of excluded addresses.

[...]
Excluded ranges:         15 ipaddresses
                        10.0.0.4 -> 10.0.0.19
[...]

Maybe here 
"def format_network_excluded_ranges(info: Dict[str, Any], padding: int = 25) -> None:"  
in 
https://github.com/unioslo/mreg-cli/blob/339857f7d3cd91e8cace132863a101317a68b778/mreg_cli/network.py#L71

terjekv:
- Always add one to the count of excluded IPs to include the last IP in the range

Fixes #196

### DIFF
--- a/ci/testsuite-result.json
+++ b/ci/testsuite-result.json
@@ -10076,7 +10076,7 @@
       "                         10.0.1.2",
       "                         10.0.1.3",
       "                         10.0.1.255 (broadcast)",
-      "Excluded ranges:         10 ipaddresses",
+      "Excluded ranges:         11 ipaddresses",
       "                         10.0.1.20 -> 10.0.1.30",
       "Used addresses:          1",
       "Unused addresses:        239 (excluding reserved adr.)"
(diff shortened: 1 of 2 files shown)


terjekv:
Tests fail as we now count the excluded list differently. The only real question is if we actually exclude that last address or not. It looks like we do: https://github.com/unioslo/mreg/blob/master/mreg/models/network.py#L199-L223

However, I would love it if code that like explicitly used the term "inclusive" or "not inclusive". Bah.

oyvindkolbu:
> However, I would love it if code that like explicitly used the term "inclusive" or "not inclusive". Bah.

The str function uses '[ .. ]' meaning inclusive. But yes, could be written explicitly in the doc string.

pederhan (approved):
LGTM, just make sure this is rebased/merged/whatever into #191 so we don't experience regressions when we merge that.

PR #199, opened 2024-01-12
Filter away timestamps from diffs, increase testing.

terjekv:
- This PR filters away timestamps from test results via a regexp.
  - The timestamps are replaced inline so other content in the same value are retained.
  - Adds testing of object history when supported (host, group, atom, role).
  - Moves filtering testing to a host.
  - Will conflict with https://github.com/unioslo/mreg-cli/pull/198, due to having to fix group history here as well.

### DIFF
--- a/ci/diff.py
+++ b/ci/diff.py
@@ -19,11 +19,6 @@ def replace_timestamps(obj: Any) -> Any:
     elif isinstance(obj, list):
         return [replace_timestamps(elem) for elem in obj]
     elif isinstance(obj, str):
-        if timestamp_pattern.match(obj):
-            print("Replacing timestamp:", obj)
-            tmp = timestamp_pattern.sub("<TIME>", obj)
-            print("With:", tmp)
-
         return timestamp_pattern.sub("<TIME>", obj)
     return obj
 
--- a/ci/diff.py
+++ b/ci/diff.py
@@ -1,11 +1,43 @@
 import difflib
 import json
+import re
 import sys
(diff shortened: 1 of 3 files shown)


ci/diff.py
```diff
@@ -1,12 +1,39 @@
 import difflib
 import json
+import re
 import sys
+from typing import Any, Dict, List
 
 
-def group_objects(json_file_path):
+def replace_timestamps(obj: Any) -> Any:
```
pederhan, line 8:
```suggestion
from typing import TypeVar

TSType = TypeVar("TSType", str, list, dict)

def replace_timestamps(obj: TSType) -> TSType:
```

This is not perfect, but provides some very rudimentary type checking:

```py
reveal_type(replace_timestamps("2020-01-01T00:00:00.000Z"))
reveal_type(replace_timestamps(["2020-01-01T00:00:00.000Z", "2020-01-01T00:00:00.000Z"]))
reveal_type(replace_timestamps({"foo": "2020-01-01T00:00:00.000Z"}))
```
```
% mypy ci
ci/diff.py:27: note: Revealed type is "builtins.str"
ci/diff.py:28: note: Revealed type is "builtins.list[Any]"
ci/diff.py:29: note: Revealed type is "builtins.dict[Any, Any]"
```

With a more complicated `TypeVar` we could remove the `Any` from the returned `list` and `dict` type annotations, but that's overkill for something that isn't even part of the CLI itself.

PR #201, opened 2024-01-19
Add list of policies to host info output.

### LINKED ISSUE #155: Add list of policies to host info output
Host info should return list of policies for a host, as that is information users often want.

terjekv:
Adds a comma separated list of policies to the end of `host info`:

```
mreg> host info foo
Name:         foo.example.org
Contact:      admin@example.org
[...]
TXT:          v=spf1 -all
Policies:     fruit, vegetable
```

In other news... We could really generalize and clean up the output commads _a lot_. :(

Fixes #155

### DIFF
--- a/ci/testsuite
+++ b/ci/testsuite
@@ -147,10 +147,12 @@ policy atom_delete apple
 policy set_description orange "Juicy"
 policy rename orange tangerine
 host add foo
+host info foo
 policy host_add tangerine foo # should fail, tangerine is an atom, not a role
 policy host_add fruit foo
 policy list_hosts fruit
 policy host_list foo
+host info foo
 policy host_remove fruit foo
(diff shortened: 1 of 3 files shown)


pederhan (approved):
LGTM

oyvindkolbu:
Not specific to this commit, but I really don't like the timestamp churn.

oyvindkolbu:
> Timestamp churn? I don't follow. :(

In the testsuite. Makes it hard to follow actual changes versus timestamp only changes.

terjekv:
> > Timestamp churn? I don't follow. :(
> 
> In the testsuite. Makes it hard to follow actual changes versus timestamp only changes.

Ah, yeah. There was a brief discussion of removing the timestamps before they hit the test results so new test results don't differ as much. That's something that should be done.

PR #202, opened 2024-01-23
Allow assoc of a MAC with 1 IPv4 + 1 IPv6 on the same VLAN by supplying hostname

### LINKED ISSUE #141: Let you assoc a MAC to a host with 1 IPv4 + 1 IPv6 by supplying hostname
When checking if a machine has several IP addresses to make an assoc here: https://github.com/unioslo/mreg-cli/blob/05011b15ff4b964acb3b2c12114e0df9d3920a32/mreg_cli/dhcp.py#L32 then both IPv4 and IPv6 addresses are counted. A machine that has one IPv4 address and one IPv6 address is then told to do an assoc on IP and not name. Should we create a special case for this situation and make it associate with the same MAC?

terjekv:
- This assumes that the VLAN IDs map between IPv4 and IPv6 networks correctly.
  - We return the IPv4 address when asking for a unique IP for a host with both IPv4 and IPv6 adresses on the same VLAN.
  - We assign the MAC to the IPv4 address when being asked to assoc against a host with both IPv4 and IPv6 on the same VLAN.
  - From a DUID perspective, this is probably the sanest thing until mreg supports DHCPv6 in a more flexible way.

Fixes #141

### DIFF
--- a/ci/testsuite
+++ b/ci/testsuite
@@ -173,7 +173,7 @@ permission network_remove 10.0.0.0/24 somegroup "[abc]+.uio.no"
 
 ##### Tests: DHCP
 #   - associate / disassociate ip with mac address
-network create -network 10.0.0.0/24 -desc foo
+network create -network 10.0.0.0/24 -desc foo -vlan 1234
 host add foo
 host a_add foo 10.0.0.5
 dhcp assoc 1.2.3.4 aa:bb:cc:dd:ee:ff  # doesn't work, because ip addr doesn't exist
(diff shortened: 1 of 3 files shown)


pederhan (approved):
LGTM

I think we should also consider refactoring `host_info_by_name()` in a future PR to make it safer to access the result. Stuff like this isn't very robust.

```py
 info["ipaddresses"][0]["ipaddress"]
```

terjekv:
Tbh the goal should be to return pydantic objects from API calls. That would alleviate the vast majority of such stuff.

pederhan:
Agreed. Being able to guarantee that the returned data actually contains a field called `ipaddresses`  which is a list of objects that each have a field called `ipaddress` is basically what I want, and Pydantic is probably the best way to achieve that.

PR #204, opened 2024-01-29
Legacy: Fix command descriptions using tuples

pederhan:
Fixes certain command descriptions being defined as tuples instead of strings.


terjekv (approved):
LGTM, thanks!

PR #206, opened 2024-01-29
Clean up authentication mess.

### LINKED ISSUE #142: Parameters are lost on reauthentication?
Noticed this weird behavior:

```
mreg> host find -name ThisDoesNotExist
You need to re-autenticate
Enter password: ******************************
WARNING: : Too many hits, 56388, more than limit of 500. Refine search.
mreg> host find -name ThisDoesNotExist
WARNING: : No hosts found.
```

Are the parameters somehow scrambled/lost when reauthenticating?

### LINKED ISSUE #160: Re-authentication interrupts action
### Issue:

When the mreg-token expires and an action is performed `mreg-cli` returns
```
WARNING: : GET "https://mreg.uio.no/api/v1/hosts/?page=2": 404: Not Found
{
  "detail": "Invalid page."
}
```
After re-authenticating again instead of fulfilling the request

### Steps to reproduce

1. Log into mreg-cli
2. Delete `.mreg-cli_auth_token`
3. Perform any action `host info bybanen`
4. Re-authenticate

### LINKED ISSUE #205: Incorrect authentication credentials doesn't prevent CLI from launching
```
mreg-cli --user pederhan --url https://mreg.uio.no --domain uio.no 
Connecting to https://mreg.uio.no
Password for pederhan: ********
ERROR: : {'detail': 'Incorrect authentication credentials.'}
Type -h for help.
mreg>
```
This used to raise an error and exit the application before #191 was merged.

On using a command, the application prompts again for new authentication information, so this isn't an _error_ per say, it's just divergent behavior from pre-#191.

terjekv:
- Cleans up the names for the auth functions.
  - Adds a LoginFailedError exception.
  - Passes on params from the pre-relogin command to the post-login command.

Fixes #205, fixes #160,  fixes #142

### DIFF
--- a/mreg_cli/exceptions.py
+++ b/mreg_cli/exceptions.py
@@ -58,3 +58,20 @@ class NetworkNotFoundWarning(CliWarning):
     """Warning class for network not found."""
 
     pass
+
+
+class LoginFailedError(CliException):
+    """Error class for login failure."""
+
+    def formatted_exception(self) -> str:
+        """Return a string formatted with 'Login failed:' prefixing the error message.
+
+        :returns: Formatted error message.
+        """
+        return f"Login failed: {super().__str__()}"
+
+    def __str__(self) -> str:
+        """Return the error message."""
+        return "Login failed"
+
+    pass
--- a/mreg_cli/utilities/api.py
+++ b/mreg_cli/utilities/api.py
@@ -23,7 +23,7 @@ if TYPE_CHECKING:
 from prompt_toolkit import prompt
 
 from mreg_cli.config import MregCliConfig
-from mreg_cli.exceptions import CliError
+from mreg_cli.exceptions import CliError, LoginFailedError
 from mreg_cli.log import cli_error, cli_warning
 
 session = requests.Session()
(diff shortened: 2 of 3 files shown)


mreg_cli/exceptions.py
```diff
@@ -58,3 +58,20 @@ class NetworkNotFoundWarning(CliWarning):
...
+    def formatted_exception(self) -> str:
+        """Return a string formatted with 'Login failed:' prefixing the error message.
+
+        :returns: Formatted error message.
+        """
+        return f"Login failed: {super().__str__()}"
+
+    def __str__(self) -> str:
+        """Return the error message."""
+        return "Login failed"
```
pederhan, line 75: Could this call `self.formatted_exception` instead for consistency?

```suggestion
        return self.formatted_exception()
```

mreg_cli/utilities/api.py
```diff
@@ -80,18 +90,29 @@ def login1(user: str, url: str) -> None:
...
     print(f"Connecting to {url}")
-    # get url
     password = prompt(f"Password for {user}: ", is_password=True)
     try:
-        _update_token(user, password)
+        auth_and_update_token(user, password)
     except CliError as e:
-        e.print_self()
+        if catch_exception:
+            e.print_self()
```
pederhan, line 113: Note: would be nice if we had unit tests that ensured we print something useful here.

I think it's fine for now, but this behavior is a bit opaque from the caller's perspective.

mreg_cli/exceptions.py
terjekv: The problem (ish) is that I'm not sure what we should say/leak to the user when logging in fails. This is why I made the default `__str__` be completely vague.

mreg_cli/utilities/api.py
terjekv: Well, mreg-cli should have unit tests. :)

mreg_cli/exceptions.py
pederhan: Sure. That makes a lot of sense from a security perspective.

pederhan (approved):
LGTM

PR #208, opened 2024-02-12
Allow oneshot commands from the terminal.

terjekv:
This patch allows the use of mreg-cli without interaction, ala:

`mreg-cli host info foo`

This will then feed the command `host info foo` to the cli and then exit.

(This is a user request from UiO)

### DIFF
--- a/mreg_cli/main.py
+++ b/mreg_cli/main.py
@@ -113,6 +113,14 @@ def main():
     )
 
     output_args.add_argument(
+        "--token-only",
+        dest="token_only",
+        action="store_true",
+        default=False,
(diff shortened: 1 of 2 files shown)


pederhan (approved):
LGTM. 

We should also look into some way to explicitly pass in credentials when running such oneshot commands, so that scripts don't get stuck on the prompt from `prompt_for_password_and_login()`. For now, this is a reasonable implementation given that users are aware of its limitations.

terjekv:
Or rather, if we're doing one shot, we should fail if there is no valid token. Otherwise we have to store the password in memory somewhere (env or whatnot).

pederhan (approved):

PR #211, opened 2024-02-19
Make `host remove` a bit safer to use.

### LINKED ISSUE #210: When deleting a host, require force if it has an MX RR or other host-bound data set.
Deleting hosts that have the RR MX set is a bad idea. One might see this in `host info`, and this should be caught:

````
MX:           Priority Server
                    10 mydomain.com
````                    

Hosts may also have other services set directly on themselves, such as DKIM, _autodiscover, or subdomain setups. Ideally this should all be caught.

(reported due to a mishap at UiO.)

### PR DESCRIPTION
**Pull request by terjekv**
## The problem...

It was much too easy to do `host remove foo -f` and have it bite in the worst possible way.

## The solution?

In this PR, `-force` is no longer a catch-all for making `host remove`, well, remove the host. If the host has cnames, ptrs, mx, srvs or ipadresses across multiple vlans, the user needs to declare a desire to override each of these explicitly via the new option `-override`.

## Supported input

Accepted overrides are `cname`, `ipadress`, `mx`, `srv`, `ptr`, `naptr`. Invalid overrides offered as parameters to `-override` are errors and the command will report on the unexpected input and stop before executing anything.

The choice to have these as textual inputs is intended with the explicit goal of making their use require more than tab-completing an option.

## Example usage

`host remove foo -force -override cname,ipaddress,mx`.

This would allow deletion / removal of a host with cnames, ipadresses across different vlans, and mx set. However, any ptr or srv RRs will still cause the deletion to cancel.

## Example warning from `host remove`

```
WARNING: : bar.example.org requires force and cnames as overrides for deletion:
  1 cnames, override with 'cname'
    - fubar.example.org
  multiple ipaddresses on the same VLAN. Must use 'force'."
```

## Notes:

1. `-force` alone works on multiple ipadresses from the same VLAN.
2. `-override` requires the presence of `-force`. The documentation states this both from expanded inline help and `host remove -h`.

Fixes #210

### DIFF
--- a/mreg_cli/commands/host_submodules/core.py
+++ b/mreg_cli/commands/host_submodules/core.py
@@ -12,7 +12,7 @@ Commands implemented:
 """
 
 import argparse
-from typing import Dict, Union
+from typing import Dict, List, Union
 
 from mreg_cli.commands.host import registry as command_registry
 from mreg_cli.exceptions import HostNotFoundWarning
@@ -140,6 +140,12 @@ def add(args: argparse.Namespace) -> None:
             metavar="NAME/IP",
         ),
         Flag("-force", action="store_true", description="Enable force."),
+        Flag(
+            "-override",
+            short_desc="Override list",
+            description="Override list for forcing (cname, naptr, srv, ptr)",
+            metavar="OVERRIDE",
+        ),
     ],
 )
 def remove(args: argparse.Namespace) -> None:
@@ -149,16 +155,27 @@ def remove(args: argparse.Namespace) -> None:
     """
     # Get host info or raise exception
     info = host_info_by_name_or_ip(args.name)
+    overrides: List[str] = args.override.split(",") if args.override else []
+
+    def forced(override_required: str = None) -> bool:
+        # If force wasn't set at all, return false.
+        if not args.force:
+            return False
+
+        if override_required:
+            return override_required in overrides
+
+        return False
 
     warn_msg = ""
     # Require force if host has any cnames.
     cnames = info["cnames"]
     if len(cnames):
-        if not args.force:
+        if not forced("cnames"):
             warn_msg += "{} cnames. ".format(len(cnames))
 
     # Require force if host has multiple A/AAAA records
-    if len(info["ipaddresses"]) > 1 and not args.force:
+    if len(info["ipaddresses"]) > 1 and not forced():
         warn_msg += "{} ipaddresses. ".format(len(info["ipaddresses"]))
 
     if len(info["mxs"]) > 0 and not args.force:
--- a/mreg_cli/commands/host_submodules/core.py
+++ b/mreg_cli/commands/host_submodules/core.py
@@ -12,7 +12,7 @@ Commands implemented:
 """
 
 import argparse
-from typing import Dict, List, Union
+from typing import Dict, List, Optional, Union
 
 from mreg_cli.commands.host import registry as command_registry
 from mreg_cli.exceptions import HostNotFoundWarning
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by pederhan — APPROVED** [review]

File: mreg_cli/commands/host_submodules/core.py [hunk 1]
```diff
@@ -140,34 +141,81 @@ def add(args: argparse.Namespace) -> None:
             metavar="NAME/IP",
         ),
         Flag("-force", action="store_true", description="Enable force."),
+        Flag(
+            "-override",
+            short_desc="Comma separated override list, requires -force.",
+            description=(
+                "Comma separated overrides for forced removal. Requires -force."
+                "Accepted overrides: 'cname', 'ipadresses', 'mxs', 'srv', 'ptr', 'naptr'."
+                "Example usage: '-override cnames,ipaddresses,mxs'"
```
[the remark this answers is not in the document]
pederhan: Should overrides all be in singular form to match the DNS record types? I.e. `cname` instead of `cnames`. Granted, ipaddress is not a record type, but that shouldn't be too confusing in this context.

mreg_cli/commands/host_submodules/core.py
```diff
@@ -140,34 +141,81 @@ def add(args: argparse.Namespace) -> None:
...
     """Remove host.
 
-    :param args: argparse.Namespace (name, force)
+    :param args: argparse.Namespace (name, force, override)
     """
     # Get host info or raise exception
     info = host_info_by_name_or_ip(args.name)
+    overrides: List[str] = args.override.split(",") if args.override else []
+
+    accepted_overrides = ["cnames", "ipaddresses", "mxs", "srvs", "ptr", "naptrs"]
```
pederhan, line 165:
```suggestion
    accepted_overrides = ["cname", "ipaddress", "mx", "srv", "ptr", "naptr"]
```

mreg_cli/commands/host_submodules/core.py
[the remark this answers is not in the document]
pederhan:
```suggestion
                "Accepted overrides: 'cname', 'ipadress', 'mx', 'srv', 'ptr', 'naptr'."
                "Example usage: '-override cname,ipaddress,mx'"
```

pederhan (commented):
Consider if overrides should be in singular form (see comments). Sorry about the ordering, I deleted and rewrote the first review comment, and now it's a bit jumbled up.

mreg_cli/commands/host_submodules/core.py
```diff
@@ -140,34 +141,81 @@ def add(args: argparse.Namespace) -> None:
...
+        warnings.append(f"  {len(info['cnames'])} cnames, override with 'cname'")
+        for cname in info["cnames"]:
+            warnings.append(f"    - {cname['name']}")
+
+    # Require force if host has multiple A/AAAA records and they are not in the same VLAN.
+    if len(info["ipaddresses"]) > 1:
+        same_vlan = ips_are_in_same_vlan([ip["ipaddress"] for ip in info["ipaddresses"]])
+
+        if same_vlan and not forced():
+            warnings.append("  multiple ipaddresses on the same VLAN. Must use 'force'.")
```
pederhan, line 194: Should this be more clear that we are referring to the option `-force`, not some argument?

```suggestion
            warnings.append("  multiple ipaddresses on the same VLAN. Must use -force.")
```

mreg_cli/commands/host_submodules/core.py
```diff
@@ -211,8 +263,11 @@ def remove(args: argparse.Namespace) -> None:
         info["ipaddress"] = info["ipaddresses"][0]["ipaddress"]
 
     # Warn user and raise exception if any force requirements was found
-    if warn_msg:
-        cli_warning("{} has: {}Must force".format(info["name"], warn_msg))
+    if warnings:
+        warn_msg = "\n".join(warnings)
+        cli_warning(
+            "{} requires force and override for deletion:\n{}".format(info["name"], warn_msg)
```
pederhan, line 269: Same thing here?

```suggestion
            "{} requires -force and -override for deletion:\n{}".format(info["name"], warn_msg)
```

pederhan (commented):
Some suggestions regarding warning messages. This _does_ add more hardcoding of option names in the messages, which isn't _great_.

PR #214, opened 2024-04-09
Migrate to pydantic

terjekv:
Another large scale rewrite to move away from a bundle of dicts into something actually structured into Pydantic models. The goal here is to be able to add features and output data without needing to test, guess, and generally fiddle with dicts everywhere...


pederhan:
I've used Pydantic for a few CLI projects by now, and there are a couple of points I think you should have in the back of your mind when developing this further:

1. How will you handle data that fails validation? When fetching _all_ hosts, should a single failing host break the entire host fetching functionality of the application, or will you implement some sort of best-effort fetching?
2. When validation fails, how do you notify the user about it? Do you leak the entire stack trace, show just the Pydantic ValidationError message, or write some custom exception message or handler that hides it from the user entirely?

terjekv:
Yeah, I've been down the same thoughts myself. Tbh, these kinds of problems are why I'm looking at gRPC for other projects, and also why they tend to come in Rust. Which doesn't really help the here and now. :p

mreg_cli/utilities/api.py
```diff
@@ -262,27 +262,170 @@ def get(
...
+
     ret: List[Dict[str, Any]] = []
-    while path:
+
+    # Get the first page to check the number of hits, and raise an exception if it is too high.
+    get_params = params.copy()
+    # get_params["page_size"] = 1
+    resp = get(path, get_params).json()
+    if "count" in resp and resp["count"] > max_hits_to_allow:
+        raise cli_warning(f"Too many hits ({resp['count']}), please refine your search criteria.")
```
pederhan, line 410:
```suggestion
        raise CliWarning(f"Too many hits ({resp['count']}), please refine your search criteria.")
```

terjekv: Yikes. Fixed.

pederhan (approved):
LGTM (again) 🤠

PR #215, opened 2024-04-23
Upgrade models.py to Pydantic V2 semantics

pederhan:


mreg_cli/api/models.py
```diff
@@ -454,13 +470,14 @@ class Role(FrozenModelWithTimestamps, APIMixin["Role"]):
 
     id: int  # noqa: A003
     created_at: datetime = Field(..., validation_alias=AliasChoices("create_date", "created_at"))
-    hosts: List[str]
-    atoms: List[str]
+    hosts: NameList
+    atoms: NameList
```
pederhan, line 474: You can decide if you want to use `List[Annotated[str, BeforeValidator(_extract_name)]]` for each field or keep using `NameList`. I suppose keeping the `NameList` type around could be useful if we have more such fields?

mreg_cli/api/models.py
```diff
@@ -617,15 +627,17 @@ class IPAddress(FrozenModelWithTimestamps, WithHost, APIMixin["IPAddress"]):
     macaddress: Optional[MACAddressField] = None
     ipaddress: IPAddressField
 
-    @validator("macaddress", pre=True, allow_reuse=True)
-    def create_valid_macadress_or_none(cls, v: str):
+    @field_validator("macaddress", mode="before")
+    @classmethod
+    def create_valid_macadress_or_none(cls, v: str) -> MACAddressField | None:
```
terjekv, line 632: `Union[MACAddressField, None]`, sadly.

mreg_cli/api/models.py
```diff
@@ -1072,13 +1085,15 @@ class Host(FrozenModelWithTimestamps, APIMixin["Host"]):
...
+    @classmethod
     def validate_name(cls, value: str) -> HostT:
         """Validate the hostname."""
         return HostT(hostname=value)
 
-    @validator("comment", pre=True, allow_reuse=True)
-    def empty_string_to_none(cls, v: str):
+    @field_validator("comment", mode="before")
+    @classmethod
+    def empty_string_to_none(cls, v: str) -> str | None:
```
terjekv, line 1096: `Union[str,None]`

mreg_cli/api/models.py
terjekv: NameList can also offer some more stuff like outputting the values etc.

mreg_cli/api/models.py
pederhan: Right. Python3.6 with no `from __future__ import annotations` :(

PR #216, opened 2024-04-24
Add `from __future__ import annotations` lint rule

pederhan:
Adds a Ruff linting rule that imports  `from __future__ import annotations` in every source file. I have also run pyupgrade in `--py311-plus` mode afterwards to rewrite string annotations.


terjekv (approved):
LGTM!

PR #217, opened 2024-05-02
Add `Atom` & `Label` models + base Host Policy model

pederhan:


mreg_cli/api/models.py
```diff
@@ -255,30 +262,58 @@ def is_delegated(self) -> bool:
...
+        elif isinstance(value, datetime):
+            return value
+        elif isinstance(value, date):
+            return datetime.combine(value, datetime.min.time())
+        return value  # let pydantic throw the ValidationError
+
+    @property
+    def created_at_with_tz(self) -> datetime:
+        """The constructed `created_at` field with timezone info from `updated_at`."""
+        return self.created_at.replace(tzinfo=self.updated_at.tzinfo)
```
pederhan, line 302: Need a better way to handle this. We can't use an after validator on the model itself due to the fields being frozen. We could assign the `created_at` field to a hidden field, and then add a `@computed_field` called `create_date` which returns a tz aware create time.

terjekv: At some point this feels like something we should fix on the server side, in the database.

pederhan: I was hoping this would work, but you cannot override normal fields with computed fields:

```py
class HostPolicyBase(FrozenModelWithTimestamps):
    # ...
    # The constructed datetime object without timezone info.
    created_at_tz_naive: datetime = Field(
        ...,
        validation_alias=AliasChoices("create_date", "created_at", "created_at_tz_naive"),
        exclude=True,
    )
    # ...

    @computed_field
    def created_at(self) -> datetime:
        """The constructed `created_at` datetime object with timezone info injected from `updated_at`."""
        return self.created_at_tz_naive.replace(tzinfo=self.updated_at.tzinfo)

    @computed_field
    def create_date(self) -> date:
        """Original `create_date` field."""
        return self.created_at_tz_naive.date()
```

```
  File "/Users/pederhan/repos/mreg-cli/venv/lib/python3.11/site-packages/pydantic/_internal/_model_construction.py", line 474, in set_model_fields
    fields, class_vars = collect_model_fields(cls, bases, config_wrapper, types_namespace, typevars_map=typevars_map)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pederhan/repos/mreg-cli/venv/lib/python3.11/site-packages/pydantic/_internal/_fields.py", line 242, in collect_model_fields
    raise ValueError("you can't override a field with a computed field")
ValueError: you can't override a field with a computed field
````

Comment (pederhan) [reply mreg_cli/api/models.py:302]: I guess we'll just do the "dumb" thing and leave it in a dangerous state and add a very stern comment about using it for comparisons.

Comment (pederhan) [reply mreg_cli/api/models.py:302]: >  You can't override a field from a parent class with a computed_field in the child class. mypy complains about this behavior if allowed, and dataclasses doesn't allow this pattern either.

_From the docstring of `computed_field`._ 

So this strictness from Pydantic's side is just a result of type-checking limitations and wanting consistency with dataclasses 🤦

Comment (terjekv) [reply mreg_cli/api/models.py:309]: Well, we could rename the field `_created_at` and then make the `@property` `created_at`? I mean... We could?

Comment (pederhan) [reply mreg_cli/api/models.py:309]: Fields can't be prefixed with underscores; Pydantic throws an exception if you try to assign `Field(...)` to an attribute name with a leading underscore. 

But yeah, I was playing with the idea of doing something like that, the only "problem" is that we have to customize model serialization with `@model_serializer` if we want to include the property if/when we serialize the model (maybe not very relevant?), hence why I was trying to make it work with `@computed_field`.

Comment (terjekv) [reply mreg_cli/api/models.py:309]: I don't really see the serialisation gain at the moment, as we're only delivering human-targeted output, and if that becomes an issue, we can deal with it then (famous last words). `create_date_not_to_be_used_stop_thinking_about_it_I_am_broken` would be an option. :)

Comment (pederhan) [reply mreg_cli/api/models.py]: Overriding a field with a `@property` when subclassing does not work. It returns the property decorator instead of the value. 

Instead of hacking together something brittle, I have changed `HostPolicy` to inherit from `FrozenModel`, and then implemented the `created_at` and `updated_at` fields directly on the model, so we can be sure this special case works. Copy-pasting `output_timestamps` is NOT ideal, but we can refactor it to use a common function if we decide to use this approach.

File: mreg_cli/api/models.py [hunk 2]
```diff
@@ -255,42 +262,134 @@ def is_delegated(self) -> bool:
...
+
+    def output(self, padding: int = 14) -> None:
+        """Output the host policy object to the console.
+
+        Subclasses should provide their own output method and call this method
+        first to output the commmon fields.
+        """
+        output_manager = OutputManager()
+        output_manager.add_line(f"{'Name:':<{padding}}{self.name}")
+        self.output_timestamps(padding=padding)
```
pederhan, line 349: This will now show `Updated:` when doing `policy info`, which it previously didn't. Remove or keep?

terjekv: I did the same for other models, I think it's nice to have. I'm thinking we can add some config options to allow users to customise some aspects of the output as well.

pederhan: Then we definitely need to unify the two `output_timestamps` methods, since I just copied it over to the new `HostPolicy` base class.

terjekv: I look forward to the day when a bit of copy and paste are the problems we fret over. :)

mreg_cli/api/models.py
```diff
@@ -255,42 +262,134 @@ def is_delegated(self) -> bool:
...
+
+        :param name: The name to search for.
+        :returns: The Atom or Role if found.
+        :raises CliWarning: If the Atom or Role is not found.
+        """
+        role_or_atom: Role | Atom | None = None
+        for func in [Atom.get_by_name, Role.get_by_name]:
+            try:
+                role_or_atom = func(name)
+            except Exception:
```
pederhan, line 329: Forgot to change this. This should definitely be `CliWarning`!

mreg_cli/api/models.py
```diff
@@ -255,42 +263,134 @@ def is_delegated(self) -> bool:
...
-        """Validate and convert the created_at field to datetime.
+    def get_by_name(cls, name: str) -> Atom | Role:
+        """Get an Atom or Role by name.
+
+        :param name: The name to search for.
+        :returns: The Atom or Role if found.
+        :raises CliWarning: If the Atom or Role is not found.
+        """
+        role_or_atom: Role | Atom | None = None
+        for func in [Atom.get_by_name, Role.get_by_name]:
```
terjekv, line 327: I'm not *completely* sold on this way of using exceptions (maybe we should have an option for getting these types that may return None), but the current implementation does work.

terjekv (approved):

mreg_cli/api/models.py
```diff
@@ -255,42 +263,134 @@ def is_delegated(self) -> bool:
...
+
+    @field_validator("created_at_tz_naive", mode="before")
+    @classmethod
+    def validate_created_at(cls, value: Any) -> datetime:
+        """Convert a datetime string to a datetime object.
+
+        :param value: The input value - should be a datetime string.
+        :returns: The input value converted to a datetime object.
+        """
+        # Fast path for str (most likely input type)
```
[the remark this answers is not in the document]
pederhan: Sure, I can add some inline comments.

mreg_cli/api/models.py
pederhan: Yes, that would be ideal. I am not overly happy catching `CliWarning`, but given the available methods it seemed to be the method involving the least boilerplate.

terjekv: Yeah, good point. We will have to do a second cleaning pass of this at some point anyway. There are more pressing issues to deal with.

mreg_cli/api/models.py
```diff
@@ -255,42 +263,134 @@ def is_delegated(self) -> bool:
...
+                role_or_atom = func(name)
+            except CliWarning:
+                pass
+        if role_or_atom is None:
+            cli_warning(f"Could not find an atom or a role with name: {name!r}")
+        return role_or_atom
 
-        :param value: Input date string from the JSON.
-        :returns: Converted datetime object.
+    def output_timestamps(self, padding: int = 14) -> None:
```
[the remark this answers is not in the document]
pederhan: Yes, I think it'll break some stuff and add a bunch of ugly workarounds, and all we get in return is avoiding some duplicate code.

mreg_cli/api/models.py
```diff
@@ -255,42 +263,134 @@ def is_delegated(self) -> bool:
...
+    @computed_field
+    def created_at(self) -> datetime:
+        """Creation time."""
+        return self.created_at_tz_naive.replace(tzinfo=self.updated_at.tzinfo)
 
-    @field_validator("created_at", mode="before")
     @classmethod
-    def validate_created_at(cls, value: str) -> datetime:
-        """Validate and convert the created_at field to datetime.
+    def get_by_name(cls, name: str) -> Atom | Role:
```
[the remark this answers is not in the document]
pederhan: Yes, I didn't really get a good grip on the naming of methods for fetching stuff.  A consistent interface for methods returning `T | None` vs `T` would be great.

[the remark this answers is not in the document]
terjekv: Totally my bad. We don't have to do anything now, but yeah.

terjekv (approved):

PR #220, opened 2024-05-07
Add policy commands

pederhan:


mreg_cli/api/models.py
```diff
@@ -523,6 +529,16 @@ def get_roles_with_atom(cls, name: str) -> list[Role]:
...
+    def add_atom(self, atom: str) -> bool:
+        """Add an atom to the role."""
+        resp = post(Endpoint.HostPolicyRoleAddAtom.with_params(self.name), name=atom)
+        return resp.ok if resp else False
+
+    def remove_atom(self, atom: str) -> bool:
+        """Remove an atom from the role."""
+        resp = delete(Endpoint.HostPolicyRoleRemoveAtom.with_params(self.name, atom))
+        return resp.ok if resp else False
+
```
pederhan, line 541: If we're not gonna be able to re-use the `create()` and `delete()` instance methods from `APIMixin`, we could consider making these classmethods.

mreg_cli/commands/policy.py
```diff
@@ -268,20 +259,11 @@ def list_atoms(args: argparse.Namespace) -> None:
...
-    info = get_list("/api/v1/hostpolicy/atoms/", params=params)
-    if info:
-        for i in info:
-            _format(i["name"], repr(i["description"]))
+    atoms = Atom.get_list_by_name_regex(args.name)
+    if atoms:
+        Atom.output_multiple_lines(atoms)
     else:
-        manager.add_line("No match")
+        OutputManager().add_line("No match")
```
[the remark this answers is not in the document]
terjekv: Is this one of those moments where we would want `if not atoms: raise EntityNotFound`? :)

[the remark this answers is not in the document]
pederhan: Very likely!

mreg_cli/api/models.py
terjekv: I like `delete()` (or "remove") to be an object method tbh, but `create` obviously can't be. :)

pederhan: I was referring to the `{add,remove}_atom` methods, but I don't think it's important either way.

terjekv: Oh, right, blind as a bat. Ignore me.

mreg_cli/commands/policy.py
```diff
@@ -366,14 +319,8 @@ def list_members(args: argparse.Namespace) -> None:
...
-    info = get_role(args.name)
-    manager = OutputManager()
-    if info["atoms"]:
-        manager.add_line("Name:")
-        for i in info["atoms"]:
-            manager.add_line(f" {i['name']}")
-    else:
-        manager.add_line("No atom members")
+    role = Role.get_by_name_or_raise(args.name)
+    role.output_atoms()
```
pederhan, line 323: Should we delegate all printing to the model like this, or does it clutter the model definition?

terjekv: I vastly prefer to delegate the printing. I'd like the managing of output and formatting to be found in as few places / files as possible.

mreg_cli/api/models.py
```diff
@@ -521,7 +646,80 @@ def get_roles_with_atom(cls, name: str) -> list[Role]:
...
+        :returns: The updated Role object.
+        """
+        label = Label.get_by_name_or_raise(label_name)
+        if label.id not in self.labels:
+            cli_warning(f"The role {self.name!r} doesn't have the label {label_name!r}")
+
+        label_ids = self.labels.copy()
+        label_ids.remove(label.id)
+        return self.patch({"labels": label_ids}, validate=False)
```
pederhan, line 706: `Role.{add,remove}_atoms` have different return types and implementations from `Role.{add,remove}_labels`. 

Should we try to _always_ have the full implementation in the models themselves, and then basically only call those methods from the command functions? So all info, warning and error messages are all printed from the model method itself?

pederhan: Keep `cli_info()` calls in commands or move those into models as well?

pederhan: Especially shoving everything into `Role.delete()` might lead to a bit of a clumsy implementation?

```python
    def delete(self) -> bool:
        """Delete the role."""
        if self.hosts:
            hosts = ", ".join(self.hosts)
            cli_error(f"Role {self.name!r} used on hosts: {hosts}")
        if super().delete():
            cli_info(f"Deleted role {self.name!r}", print_msg=True)
        else:
            cli_error(f"Failed to delete role {self.name!r}")
        return True # a bit contrived?
```

terjekv: I'd like to redo `cli_info` to either use `OutputManager.add_ok()` or something similar, and with that reimplement logging properly (if we want logging in the client). I was thinking of looking at this when I already started on https://github.com/unioslo/mreg-cli/pull/225

pederhan: Updated my comment just now btw

terjekv: > The rule for models could be that they never print, but may raise?

This still maps to being "cleaner and more correct with regards to separation of concern"? It's at least an easy rule to follow. :)

mreg_cli/api/abstracts.py
```diff
@@ -384,7 +384,10 @@ def patch(self, fields: dict[str, Any], validate: bool = True) -> Self:
         :param validate: Whether to validate the patched object.
         :returns: The object refetched from the server.
         """
-        patch(self.endpoint().with_id(self.id_for_endpoint()), **fields)
+        if use_json:
+            patch(self.endpoint().with_id(self.id_for_endpoint()), fields, use_json=True)
+        else:
+            patch(self.endpoint().with_id(self.id_for_endpoint()), use_json=False, **fields)
```
pederhan, line 390: We need to pass in the `fields` argument as the `params` argument to `patch()` when using `use_json=True`, hence the need for differentiating them here...

mreg_cli/commands/policy.py
```diff
@@ -5,13 +5,14 @@
 import argparse
 from typing import Any
 
+from mreg_cli.api.history import HistoryResource
+from mreg_cli.api.history import HistoryResource
```
terjekv: Duplicate?

mreg_cli/api/abstracts.py
pederhan: Always prior to this it seems? `patch` was called like this under the hood:

```py
requests.patch(path, data={"field": new_value"})
```

As opposed to this (used only when removing labels from what I can tell):

```py
requests.patch(path, json={"field": new_value"})
```

🤷

mreg_cli/commands/policy.py
pederhan: Some import auto sorting weirdness when selecting ranges to commit...

mreg_cli/api/models.py
```diff
@@ -656,7 +656,7 @@ def remove_atom(self, atom_name: str) -> bool:
             if atom_name == atom:
                 break
         else:
-            cli_warning(f"Atom {atom_name!r} not a member of {self.name!r}")
+            raise EntityNotFound(f"Atom {atom_name!r} not a member of {self.name!r}")
```
pederhan, line 659: Probably need a different exception type than this? `EntityNotFound` maps to 404 in my mind, but I'm not sure what is a good replacement? Struggling to think of a good name.

mreg_cli/api/abstracts.py
terjekv: ... Well, that looks like something we probably should... fix. :D

mreg_cli/api/models.py
terjekv: I used `EntityOwnershipMismatch` for this kind of stuff.

PR #221, opened 2024-05-08
Table model

pederhan:
The idea is that we can guard a little more against human error when calling `OutputManager().add_formatted_table()`, as well as providing a minimum of type safety.

```py
from mreg_cli.tables import TableRow

class SomeTableRow(TableRow):
    hosts: str # header: Hosts
    stuff_to_add: str
    """Stuff to Add""" # Use attribute docstrings to specify custom headers
    other_thing: str = Field(..., description="Can also use field description to specify header")
    
rows = [
    SomeTableRow(
        hosts="foo.example.com",
        stuff_to_add="stuff",
        other_thing="thing",
    )
]

OutputManager.add_formatted_table(
    headers=SomeTableRow.headers(),
    keys=SomeTableRow.keys(),
    data=rows,
)
```

### DIFF
--- a/mreg_cli/tables.py
+++ b/mreg_cli/tables.py
@@ -1,0 +1,37 @@
+from pydantic import BaseModel, ConfigDict
+
+
+class TableRow(BaseModel):
+    """Base class for table rows used by `OutputManager.print_formatted_table`.
+
+    The Header for each column is determined by the name of each field.
+    However, names can be overridden by specifying an attribute docstring
+    for the field _or_ using `Field(..., description="<header>")`.
+
+    Example:
+    -------
+    ```
+    class MyRow(TableRow):
+        foo: str # header: "Foo"
+        bar: str # header: "Bar Users"
+        \"\"\"Bar Users\"\"\"
+        baz: str = Field(..., description="Baz Users") # header: "Baz Users"
+    ```
+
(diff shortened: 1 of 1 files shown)


terjekv:
Oooh, I like the way this is going! Although, if we're looking at doing proper revamping of the output, I'm tempted to use [rich](https://github.com/Textualize/rich) [docs](https://rich.readthedocs.io/en/stable/index.html). That would give us tables (and much much more) with a lot of options for output, and we don't have to pretend to give two bleeps about terminals. 

We should probably also make the syntax for defining a table structure a bit more declarative and ideally reusable, so we can say "this is a table for purpose X" and use that consistently?

pederhan:
+1 to Rich tables. I've done something similar in the new version of Zabbix-CLI I am working on: https://github.com/pederhan/zabbix-cli/blob/5cff11c9623232cec44d1a1046613d0aadbb06ce/zabbix_cli/models.py#L81-L245

But I think it's a bit too complex and I'm not super happy with using `json_schema_extra` to customize columns.

terjekv:
Sounds good, and I can only give sympathy for not fixing a cli once, but twice, in similar fashion. Thank you for doing this!

pederhan:
Instead of working on this jerry-rigged solution, we should look at a more holistic approach (rich tables, nicer abstractions, etc.).

PR #223, opened 2024-05-08
dhcp commands and exception usage.

terjekv:
- Migrate more files to use exceptions and not `cli_*`-commands.
  - Migrate commands for `dhcp assoc` and `dhcp disassoc`.

There is a fair bit of refactoring to utilise all the flexibility now on offer, while retaining the safety previously desired.

### DIFF
--- a/mreg_cli/exceptions.py
+++ b/mreg_cli/exceptions.py
@@ -68,6 +68,36 @@ class DeleteFailure(CliError):
     pass
 
 
+class GetFailure(CliError):
+    """Error class for failed retrieval."""
+
+    pass
+
+
+class InternalError(CliError):
+    """Error class for internal errors."""
+
+    pass
+
+
+class APIError(CliError):
+    """Error class for API errors."""
+
+    pass
+
+
+class UnexpectedAPIData(APIError):
+    """Error class for unexpected API data."""
+
+    pass
+
+
+class ValidationFailure(CliWarning):
+    """Warning class for validation failures."""
+
+    pass
+
+
 class EntityNotFound(CliWarning):
     """Warning class for an entity that was not found."""
 
@@ -76,6 +106,12 @@ class EntityNotFound(CliWarning):
 
 class EntityAlreadyExists(CliWarning):
     """Warning class for an entity that already exists."""
+
+    pass
+
+
+class MultipleEntititesFound(CliWarning):
+    """Warning class for multiple entities found."""
 
     pass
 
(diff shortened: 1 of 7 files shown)


mreg_cli/exceptions.py
```diff
@@ -98,6 +134,30 @@ class ForceMissing(CliWarning):
...
+
+    pass
+
+
+class IsNotANetwork(CliWarning):
+    """Warning class for an entity that is not a network."""
+
+    pass
+
+
```
pederhan, line 160: nit: Non-standard naming scheme for these exception types

Presenting some alternatives below.

## `Invalid` prefix instead of `IsNot`

```suggestion
class InvalidIPAddress(CliWarning):
    """Warning class for an entity that is not an IP address."""

    pass


class InvalidIPv4Address(CliWarning):
    """Warning class for an entity that is not an IPv4 address."""

    pass


class InvalidIPv6Address(CliWarning):
    """Warning class for an entity that is not an IPv6 address."""

    pass


class InvalidNetwork(CliWarning):
    """Warning class for an entity that is not a network."""

    pass


```
##  `Error` suffix without `IsNot` prefix

```suggestion
class IPAddressError(CliWarning):
    """Warning class for an entity that is not an IP address."""

    pass


class IPv4AddressError(CliWarning):
    """Warning class for an entity that is not an IPv4 address."""

    pass


class IPv6Error(CliWarning):
    """Warning class for an entity that is not an IPv6 address."""

    pass


class NetworkError(CliWarning):
    """Warning class for an entity that is not a network."""

    pass


```


## `Invalid` prefix _and_ `Error` suffix

```suggestion
class InvalidIPAddressError(CliWarning):
    """Warning class for an entity that is not an IP address."""

    pass


class InvalidIPv4AddressError(CliWarning):
    """Warning class for an entity that is not an IPv4 address."""

    pass


class InvalidIPv6Error(CliWarning):
    """Warning class for an entity that is not an IPv6 address."""

    pass


class InvalidNetworkError(CliWarning):
    """Warning class for an entity that is not a network."""

    pass


```

## Original names with `Error` suffix:

```suggestion
class IsNotAnIPAddressError(CliWarning):
    """Warning class for an entity that is not an IP address."""

    pass


class IsNotAnIPv4AddressError(CliWarning):
    """Warning class for an entity that is not an IPv4 address."""

    pass


class IsNotAnIPv6AddressError(CliWarning):
    """Warning class for an entity that is not an IPv6 address."""

    pass


class IsNotANetworkError(CliWarning):
    """Warning class for an entity that is not a network."""

    pass


```

mreg_cli/exceptions.py
```diff
@@ -68,6 +68,36 @@ class DeleteFailure(CliError):
     pass
 
 
+class GetFailure(CliError):
```
pederhan, line 71: Inconsistent naming scheme? Is there a semantic difference in how exceptions named `*Failure` and `*Error` are used? If yes, it should be probably documented, if not shouldn't we just use `Error` everywhere?

mreg_cli/exceptions.py
```diff
@@ -68,6 +68,36 @@ class DeleteFailure(CliError):
...
+    pass
+
+
+class UnexpectedAPIData(APIError):
+    """Error class for unexpected API data."""
+
+    pass
+
+
+class ValidationFailure(CliWarning):
```
pederhan, line 95: ditto, but I understand not wanting to shadow Pydantic's `ValidationError` here.

pederhan (approved):

PR #227, opened 2024-05-14
Zone commands

pederhan:


mreg_cli/api/models.py
```diff
@@ -532,20 +534,45 @@ def create_delegation(
...
         cls = Delegation.type_by_zone(self)
-
-        resp = get(cls.endpoint_with_id(self, name))
+        resp = get(cls.endpoint_with_id(self, name), ok404=True)
         if not resp:
-            raise EntityNotFound(f"Could not find delegation {name!r} in zone {name!r}")
+            return None
         return cls.model_validate_json(resp.text)
 
+    def get_delegation_or_raise(self, name: str) -> ForwardZoneDelegation | ReverseZoneDelegation:
```
pederhan, line 550: We need to add this as far as I understand because `Delegation.get_by_name_{and,or}_raise()` doesn't work when the endpoint has 2 parameters/path components it needs to replace. I.e.: `/api/v1/zones/forward/{}/delegations/{}`.

pederhan:
https://github.com/unioslo/mreg-cli/blob/a828f88b22c474f61695e749224ae829aa143789/mreg_cli/api/endpoints.py#L125-L129

If we go through with my refactoring of the `Endpoint.with_*` methods, we should consider moving this check to `_request_wrapper`.

mreg_cli/api/models.py
pederhan: Check point 2 of https://github.com/unioslo/mreg-cli/pull/227#issuecomment-2112481006 for more information.

mreg_cli/api/models.py
```diff
@@ -386,8 +399,400 @@ def is_delegated(self) -> bool:
...
+        return zone_t.get_by_name_or_raise(name)
+
+    @classmethod
+    def get_zone_and_raise(cls, name: str) -> None:
+        """Get a zone by name, and raise if found.
+
+        :param name: The name of the zone to get.
+        """
+        zone_t = cls.type_by_name(name)
+        return zone_t.get_by_name_and_raise(name)
```
pederhan, line 582: Because `ForwardZone` and `ReverseZone` inherit from this class (`Zone`), we can't call these methods `get_by_name*`, since that would shadow the `get_by_name*` methods inherited from `WithName`, hence the different naming scheme here...

PR #231, opened 2024-05-19
Add utility functions for setting/getting auth header

pederhan:
Interface for setting/getting auth token header value.


terjekv (approved):
LGTM, much better to abstract away internals like this.

PR #235, opened 2024-05-21
Use JSON encoding for all `PATCH` requests

pederhan:
This PR removes `use_json` from `utilities.api._request_wrapper`, and makes it so _all_ `PATCH` requests are sent with JSON encoded data.


terjekv (approved):
Please and thank you. :)

PR #236, opened 2024-05-21
Fix/standardize some patch commands

pederhan:
This PR fixes/standardizes the behavior of some patch commands discovered when developing #235.

### DIFF
--- a/mreg_cli/api/models.py
+++ b/mreg_cli/api/models.py
@@ -1598,7 +1598,11 @@ class IPAddress(FrozenModelWithTimestamps, WithHost, APIMixin):
 
         :returns: A new IPAddress object fetched from the API with the MAC address removed.
         """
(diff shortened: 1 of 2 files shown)


mreg_cli/api/models.py
```diff
@@ -1598,7 +1598,11 @@ def disassociate_mac(self) -> IPAddress:
 
         :returns: A new IPAddress object fetched from the API with the MAC address removed.
         """
-        return self.patch(fields={"macaddress": ""})
+        # Model converts empty string to None so we must validate this ourselves.
+        patched = self.patch(fields={"macaddress": ""}, validate=False)
```
pederhan, line 1602: Should look into having `""` equal `None` when validating perhaps, but I fear that could create some hard to track down bugs unless we are certain this behavior is consistent in the API (which seems unlikely?!).

PR #237, opened 2024-05-21
Add `WithHistory` + fetching history for deleted objects

pederhan:
Supersedes #230

### DIFF
--- a/mreg_cli/api/models.py
+++ b/mreg_cli/api/models.py
@@ -5,20 +5,23 @@ from __future__ import annotations
 import ipaddress
 import re
 from datetime import date, datetime
-from typing import Any, Literal, Self, cast
+from typing import Any, ClassVar, Literal, Self, cast
 
 from pydantic import (
     AliasChoices,
     BaseModel,
+    ConfigDict,
     Field,
     computed_field,
     field_validator,
     model_validator,
 )
+from typing_extensions import Unpack
 
 from mreg_cli.api.abstracts import APIMixin, FrozenModel, FrozenModelWithTimestamps
 from mreg_cli.api.endpoints import Endpoint
 from mreg_cli.api.fields import IPAddressField, MACAddressField, NameList
+from mreg_cli.api.history import HistoryItem, HistoryResource
 from mreg_cli.config import MregCliConfig
 from mreg_cli.exceptions import (
     CliWarning,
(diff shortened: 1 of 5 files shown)


mreg_cli/api/models.py
```diff
@@ -334,6 +337,44 @@ def rename(self, new_name: str) -> Self:
...
+
+    def __init_subclass__(cls, **kwargs: Unpack[ConfigDict]):
+        """Ensure that subclasses implement the history_resource class var."""
+        # NOTE: Only works for Pydantic model subclasses!
+        for attr in getattr(cls, "__class_vars__", []):
+            if getattr(cls, attr) == ClassVarNotSet:
+                raise NotImplementedError(
+                    f"Subclass {cls.__name__} must implement abstract class var `{attr}`."
+                )
+        return super().__init_subclass__(**kwargs)
```
pederhan, line 364: We could consider using this pattern for `APIMixin.endpoint`, so that we ensure on startup that a class implements an endpoint. If we want to do that, we should move this check into some sort of ABC base class we can share across all models.

terjekv: I mean, this is brilliant from a code quality point of view, but the implementation is not exactly trivial. I'm assuming the internal  python APIs being used here are stable? (Well, "stable", this is python after all)...

mreg_cli/commands/host_submodules/core.py
```diff
@@ -494,4 +494,5 @@ def history(args: argparse.Namespace) -> None:
 
     :param args: argparse.Namespace (name)
     """
-    Host.get_by_any_means_or_raise(args.name).output_history(HistoryResource.Host)
+    hostname = HostT(hostname=args.name)
```
terjekv, line 497: Nice catch to ensure FQDN for the host. :)

mreg_cli/api/models.py
pederhan: `__init_subclass__` was added in 3.6, and a lot of libraries rely on it. However, after playing with this for a bit, it's _pretty_ hard to actually make something that works with multiple levels of inheritance and is also type safe. Consider this:

```py
from typing import Any, ClassVar, final

from mreg_cli.api.history import HistoryResource


ClassVarNotSet = object()


def AbstractClassVar() -> Any:
    """Hack to implement an abstract class variable on a Pydantic model."""
    return ClassVarNotSet


class ABCModel(BaseModel, ABC):
    """Pydantic model that supports abstract class variables."""

    __final__: bool = False

    def __init__(self, **data: Any) -> None:
        """Initialize the model."""
        if not self.__final__:
            raise TypeError(f"Abstract class {self.__class__.__name__} cannot be instantiated.")
        super().__init__(**data)

    def __init_subclass__(cls, **kwargs: Unpack[ConfigDict]):
        """Ensure that subclasses implement the history_resource class var."""
        if not issubclass(cls, BaseModel):
            raise TypeError("Subclasses must be Pydantic models.")

        if cls.__final__:
            for attr in cls.__class_vars__:
                if getattr(cls, attr) == ClassVarNotSet:
                    raise NotImplementedError(
                        f"Subclass {cls.__name__} must implement abstract class var `{attr}`."
                    )
        return super().__init_subclass__(**kwargs)



class WithHistory(ABCModel):
    history_resource: ClassVar[Any] = AbstractClassVar()


class HostPolicy(WithHistory):
    def do_policy_stuff(self):
        print("doing policy stuff")


@final
class Role(HostPolicy):
    history_resource: ClassVar[HistoryResource] = HistoryResource.HostPolicy_Role


@final
class Atom(HostPolicy):
    history_resource: ClassVar[HistoryResource] = HistoryResource.HostPolicy_Atom


```

It works, but here's the problem: `HostPolicy.history_resource` can be accessed (and will be set to `ClassVarNotSet`), and I don't see a good way to prevent that while also allowing subclassing. It seems that would be something we would have to resolve after all the classes have been defined, and that adds an unmanageable level of complexity.

It also necessitates marking classes as final when using this base class, as otherwise we cannot determine which classes to enforce the abstract classvar check on (from my understanding at least).

pederhan: I don't really see a good way to safely implement nested subclassing unless we resolve this _after_ all classes have been defined, which adds a TON of complexity. The way I see it, this gives us 2 options:

1. Add `WithHistory` as-is, which necessitates being added on the final subclass.
2. Add `ABCModel`, which requires classes to be marked as final in order to be checked on runtime.

And of course the 3rd, and less realistic, one:

3. Add checks that are fired _after_ models have been defined which checks the bases of all models to determine whether or not a model is final (if we find no subclass of a model, it's final).

PR #238, opened 2024-05-22
Network commands

pederhan:


mreg_cli/utilities/api.py
```diff
@@ -394,7 +395,7 @@ def _check_expect_one_result(
     get_params = params.copy()
     # get_params["page_size"] = 1
     resp = get(path, get_params).json()
-    if "count" in resp and resp["count"] > max_hits_to_allow:
+    if max_hits_to_allow and resp.get("count", 0) > abs(max_hits_to_allow):
```
pederhan, line 398: Required in order to fetch _all_ networks in `network create`.

terjekv: Yeah, as much as I don't like default of allow all. Anyway, this should also be used to resolve https://github.com/unioslo/mreg-cli/blob/cd156d189ebcc4bae7f6a6986912f6796273c60b/mreg_cli/commands/permission.py#L55-L57 

(Pardon my hackery.)

mreg_cli/types.py
```diff
@@ -47,6 +58,11 @@ class RecordingEntry(TypedDict):
 NargsType = int | NargsStr
 
 
+JSONPrimitive = Union[str, int, float, bool, None]
+JSONValue = Union[JSONPrimitive, Mapping[str, "JSONValue"], Sequence["JSONValue"]]
+JSONMapping = Mapping[str, JSONValue]
```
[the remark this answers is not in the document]
terjekv: `Union[Union[str, int, float, bool, None], Mapping[str, "JSONValue"], Sequence["JSONValue"]]` makes my eyes, nose, head, and hair hurt. I am however in awe of your ability to express Python types...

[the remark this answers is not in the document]
pederhan: Also, we don't strictly _need_ `JSONMapping`, since that is part of the `JSONValue` type, but it could be a very useful shorthand for most places where we accept some sort of JSON data in mapping form. Especially when it comes to avoiding problems stemming from invariant/covariant value types:

```py
def accepts_json_mapping(mapping: Mapping[str, JSONValue]) -> None:
    print(mapping)

def accepts_json_dict(mapping: dict[str, JSONValue]) -> None:
    print(mapping)


d1: dict[str, JSONValue] = {"a": 1, "b": "c", "d": {"e": 2.0}}
accepts_json_mapping(d1)  # OK 
accepts_json_dict(d1)  # OK 


d2: Mapping[str, JSONValue] = {"a": 1, "b": "c", "d": {"e": 2.0}}
accepts_json_mapping(d2)  # OK 
accepts_json_dict(d2)  # error: Invariant ("int | str | dict[str, float]" is not the same as "JSONValue")


d3: dict[str, str] = {"a": "b"}
accepts_json_mapping(d3)  # OK 
accepts_json_dict(d3)  # error: Invariant ("str" is not the same as "JSONValue")


d4: Mapping[str, str] = {"a": "b"}
accepts_json_mapping(d4)  # OK 
accepts_json_dict(d4)  # error: Invariant ("str" is not the same as "JSONValue")
```

This illustrates how much easier we make it for ourselves by using `Mapping` annotations where we have complex types, since something like `dict[str, str]` cannot be passed to a function annotated with `dict[str, JSONValue]`, while it can if the annotation is `Mapping[str, JSONValue]`.

I know this from (painful) first-hand experience, not because of some deep understanding of the type system. 🥲

mreg_cli/utilities/api.py
pederhan: > Yeah, as much as I don't like default of allow all. Anyway, this should also be used to resolve

The default for `max_hits_to_allow` is still 500, it's just able to handle `None` arguments now to disable the limit where needed.

mreg_cli/types.py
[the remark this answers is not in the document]
terjekv: The invariant primitives like `dict` and `list`, are weird like this. It's a similar issue where `list[Child]` isn't an acceptable type declaration for something that expects `list[Parent]`, but it works with `Sequence[Child]` and `Sequence[Parent]`. It's... frustrating.

mreg_cli/utilities/api.py
terjekv: Ah, right, explicit None, that makes a lot of sense. Thanks!

mreg_cli/commands/network.py
```diff
@@ -314,19 +269,14 @@ def find(args: argparse.Namespace) -> None:
...
-    if ip_arg:
-        ip_range = get_network_range_from_input(ip_arg)
-        network_info = get_network(ip_range)
-        if not network_info:
-            cli_warning(f"No network found for ip {ip_arg}")
-        networks = [network_info]
+    if ip_arg := args_dict.get("ip"):
+        # We technically support ID args when calling this, but that's fine.
+        net = Network.get_by_any_means_or_raise(ip_arg)
+        networks = [net]
```
pederhan, line 277: The question is: Why did this support network addresses in the first place? We should probably just remove the support for that and change this to `Network.get_by_ip_or_raise`.

mreg_cli/utilities/api.py
pederhan: > Yeah, as much as I don't like default of allow all. Anyway, this should also be used to resolve
> 
> https://github.com/unioslo/mreg-cli/blob/cd156d189ebcc4bae7f6a6986912f6796273c60b/mreg_cli/commands/permission.py#L55-L57
> 
> (Pardon my hackery.)

Fixed in [1d44a23](https://github.com/unioslo/mreg-cli/pull/238/commits/1d44a2325e2b7e6e0c9c081a5fe9e39d5d927007)

mreg_cli/commands/network.py
pederhan: Network and ID arg support removed in [f5642ff](https://github.com/unioslo/mreg-cli/pull/238/commits/f5642ff6ca524b28dd2d41c4e7867f1893ef5215)

PR #240, opened 2024-05-24
Add pagination support for `get_typed`

pederhan:
Also renames `max_hits_to_allow` to `limit` for consistency across all methods that fetch paginated results.

### DIFF
--- a/mreg_cli/utilities/api.py
+++ b/mreg_cli/utilities/api.py
@@ -433,6 +433,7 @@ def get_typed(
     :param type_: The type to which the response data should be deserialized.
     :param params: The parameters to pass to the API endpoint.
     :param paginated: Whether the response is paginated.
+    :param limit: The maximum number of hits to allow for paginated responses.
 
     :raises ValidationError: If the response cannot be deserialized into the given type.
 
--- a/mreg_cli/utilities/api.py
+++ b/mreg_cli/utilities/api.py
@@ -253,24 +253,24 @@ def get_list(
     path: str,
     params: dict[str, Any] | None = None,
     ok404: bool = False,
-    max_hits_to_allow: int | None = 500,
+    limit: int | None = 500,
 ) -> list[dict[str, Any]]:
     """Make a get request that produces a list.
 
     Will iterate over paginated results and return result as list. If the number of hits is
-    greater than max_hits_to_allow, the function will raise an exception.
+    greater than limit, the function will raise an exception.
 
     :param path: The path to the API endpoint.
     :param params: The parameters to pass to the API endpoint.
     :param ok404: Whether to allow 404 responses.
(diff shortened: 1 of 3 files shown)


mreg_cli/utilities/api.py
```diff
@@ -426,14 +432,20 @@ def get_typed(path: str, type_: type[T], params: dict[str, Any] | None = None) -
...
+    :param limit: The maximum number of hits to allow for paginated responses.
 
     :raises ValidationError: If the response cannot be deserialized into the given type.
 
     :returns: An instance of `type_` populated with data from the response.
     """
-    resp = get(path, params=params)
     adapter = TypeAdapter(type_)
-    return adapter.validate_json(resp.text)
+    if paginated:
```
terjekv, line 443: One thing that struck me. If the type is something iterable (sequence, list, etc), should we then automatically turn on pagination, rather than needing it to be set by the user?

terjekv (approved):

mreg_cli/utilities/api.py
pederhan: Will that always be correct? Is `get_list` able to handle endpoints that return results that are not encapsulated the way paginated content is` i.e.:

```jsonc

{
    "count": 1,
    "next": null,
    "previous": null,
    "results": [{"foo": "bar"]
}
```
vs

```json
[{"foo": "bar"]
```

pederhan: Nope:

```
get_list('/api/v1/networks/172.22.4.0/22/unused_list')

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/pederhan/repos/mreg-cli/mreg_cli/utilities/api.py", line 273, in get_list
    ret = get_list_generic(path, params, ok404, limit, expect_one_result=False)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pederhan/repos/mreg-cli/mreg_cli/utilities/api.py", line 398, in get_list_generic
    if limit and resp.get("count", 0) > abs(limit):
                 ^^^^^^^^
AttributeError: 'list' object has no attribute 'get'
```

pederhan: I can try to cook up something

pederhan: Maybe implemented in b4e7e5a373b7acde7afb5648514abd7e285dd83b, but not sure yet if this has broken any other `get_list` or `get_typed` calls 🤪 . Will find out on Monday.

terjekv: Nice catch. Bah. You know, with some of the legacy code around feel like we're working on painting the Sixteenth Chapel using crayons on a foundation of wet cardboard...

PR #242, opened 2024-05-27
Fix broken command(s) in CI

pederhan-uio:
Aiming to just fix all commands that are broken in CI because of incorrect `argparse.Namespace` attribute access, and other similar surface-level issues (such as any new ones created by #240).


pederhan:
THANKS VSCODE FOR USING MY GITHUB COPILOT ACCOUNT FOR THIS!

PR #243, opened 2024-05-27
Fix broken command(s) in CI

pederhan:
Aiming to just fix all commands that are broken in CI because of incorrect `argparse.Namespace` attribute access, and other similar surface-level issues (such as any new ones created by https://github.com/unioslo/mreg-cli/pull/240).


pederhan:
Just merging here, since another PR will handle updating the test suite with a more comprehensive warning/error logging regime.

oyvindhagberg (approved):

PR #239, opened 2024-05-24
Define metadata and requirements in pyproject.toml

pederhan:
Removes `setup.py` and `setup.cfg`.

PR #244, opened 2024-05-27
Require SOCKS dependencies

pederhan:
As mentioned in #239, adding SOCKS support is important for people that want to use MREG CLI from outside the network. Furthermore, since we are aiming to build binaries for this application, it's important that this is a required dependency so it's added to the binary release.


pederhan:
Unclear why this broke the pyinstaller build.

terjekv (approved):
Good to have, I suppose. :)

PR #245, opened 2024-05-28
Fix DHCP commands

pederhan:


mreg_cli/commands/dhcp.py
```diff
@@ -110,15 +105,24 @@ def disassoc(args: argparse.Namespace) -> None:
...
+                ) from None
 
             ipaddress_to_use = ips_with_mac[0]
+    else:
+        ipobjs = IPAddress.get_by_ip(addr.address)
+        if not ipobjs:
+            raise InputFailure(f"IP address {args.name} does not exist.")
+        elif len(ipobjs) > 1:
+            raise InputFailure(f"IP {args.name} is in use by {len(ipobjs)} hosts.")
+        ipaddress_to_use = ipobjs[0]
```
pederhan, line 122: This is not great! Nested try/excepts is rarely a good sign

pederhan, line 122: This is not great! Nested try/excepts is rarely a good sign

mreg_cli/commands/dhcp.py
```diff
@@ -110,15 +105,24 @@ def disassoc(args: argparse.Namespace) -> None:
...
 
-    mac = ipaddress_to_use.macaddress
     ipaddress_to_use.disassociate_mac()
+    mac = ipaddress_to_use.macaddress
```
pederhan, line 125: This is not great! Nested try/excepts is rarely a good sign

terjekv:
`get_host_by_any_means` allows for fetching by IP, but the problem here is that we need the exact IP for use later. We could do a fetch via `get_host_by_any_means` and then cycle through the IPs if we were given an IP (or a Mac actually, same issue arises then).

mreg_cli/commands/dhcp.py
```diff
@@ -85,41 +101,34 @@ def disassoc(args: argparse.Namespace) -> None:
...
+                raise InputFailure(
+                    f"Host {host} has multiple IP addresses with MAC addresses."
+                ) from None
 
-            ipaddress_to_use = ips_with_mac[0]
+            ipaddress = ips_with_mac[0]
 
-    mac = ipaddress_to_use.macaddress
-    ipaddress_to_use.disassociate_mac()
+    ipaddress.disassociate_mac()
```
terjekv, line 129: Nitpick. I always thought it looked weird to first something like `disassociate_mac()` and then fetch it via `.macaddress`.

mreg_cli/api/fields.py
```diff
@@ -47,6 +47,11 @@ class IPAddressField(FrozenModel):
 
     address: IP_AddressT
 
+    @classmethod
+    def from_string(cls, address: str) -> IPAddressField:
```
terjekv, line 51: We should migrate to this everywhere. Very good idea!

terjekv (approved):

mreg_cli/commands/dhcp.py
```diff
@@ -23,6 +24,25 @@ def __init__(self, cli: Any) -> None:
         super().__init__(cli, command_registry, "dhcp", "Manage DHCP associations.", "Manage DHCP")
 
 
+def ipaddress_from_ip_arg(arg: str) -> IPAddress | None:
```
terjekv, line 27: Add a line about what can be raised? Also use EntityOwnershipMismatch when multiple hosts are found to use the IP? Something like:

```
:raises: InputFailure when the IP does not exist.
:raises: EntityOwnershipMismatch when the IP is in use by multiple hosts.
```

mreg_cli/commands/dhcp.py
pederhan: I've removed this assignment altogether. We can access it directly when constructing the message.

PR #249, opened 2024-05-29
Strip None values when sending JSON

pederhan:
The API is unable to handle `None` values in request data. Requests removes form data with `None` values, but the same doesn't happen automatically for JSON, so we have to do it ourselves.


terjekv (approved):
I wish we didn't have to do this, but yeah.

PR #251, opened 2024-05-30
Add pytest unit tests

pederhan:
This PR adds Pytest unit tests which aim to complement the test suite by testing parts of the code that the integrations tests don't cover well (or at all). 

## Refactored `TokenFile`

The `TokenFile` class has been refactored to make it easier to test. This includes making certain methods public, and changing some methods from class methods to instance methods. 

Furthermore, a `TypeAdapter` has been added to validate the contents of the token file. Following the precedent set by the existing methods, failed validation only prints an error instead of raising an exception.

## Inline Snapshots

In order to save time on writing tests, the [`inline-snapshot`](https://github.com/15r10nk/inline-snapshot) package has been used to generate snapshots of values we test against. The usage of this should be documented, but the TL;DR is:

### When writing a new test

```
pytest --inline-snapshot=review
```

One can also skip the diff view and create snapshots without previewing the changes:

```
pytest --inline-snapshot=create
```

*NOTE: It's recommended to always use `--inline-snapshot=review` to see a diff of any modifications.*

### Fixing failing tests

To update a test after changing something that causes it to fail:

```
pytest --inline-snapshot=fix

# OR to interactively fix:

pytest --inline-snapshot=review
```


terjekv (approved):
Great work, thanks!

PR #253, opened 2024-05-30
Show git commit in `--version`

pederhan:
When mreg-cli is installed from git:

```
$ mreg-cli --version
mreg-cli version 1.0.0 (git: 4a35efeb4b738b6242345a2e169ac3e109317756)
```

NOTE: if you have an existing editable mreg-cli installation, make sure to re-install with `pip install -U -e .` to generate the version file.


pederhan:
I will handle building and publishing tagged versions in a future PR.

PR #255, opened 2024-05-31
Make testsuite work again

oyvindhagberg:
What do you think about these changes?
- Remove log.py, replace cli_info with OutputManager.add_ok and cli_warning/error with raise CliWarning/Error
- Update workflow action versions
- Add the new label commands to testsuite.json so the tests don't complain about commands being different
- Add a workaround for a server that crashes when being supplied json data in post/patch requests
- Replace formatted datetime strings in output (from @pederhan)

Remaining work:
- Still have to look through the diff between testsuite.json and the new output and verify that the changes are in fact how we want them. Looks like reformatted output messages, differences in api call url paths, and such.


terjekv (changes requested):
Some imports need cleaning and I'd like to specialise all exceptions so we don't call CliWarning and CliError directly. https://github.com/unioslo/mreg-cli/pull/256 proposes a solution to both issues. 

We should also go through the utilities and cull them, as the vast majority of those functions are never called any more. Beyond the validators and api, I can't see anything else being used any more.

ci/diff.py
```diff
@@ -1,25 +1,36 @@
...
 
+timestamp_pattern = re.compile(
+    r"\d{4}-\d{2}-\d{2}[ T]\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:[+-]\d{2}:\d{2})?|\d{4}-\d{2}-\d{2}"
+)
+datetime_str_pattern = re.compile(
+    r"\b[A-Za-z]{3}\s[A-Za-z]{3}\s([0-2]?[0-9]|3[0-1])\s([0-1]?[0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])\s[0-9]{4}\b"
+)
+# datetime_str_pattern = re.compile(
+#     r"\b(Sun|Mon|Tue|Wed|Thu|Fri|Sat)\s(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)\s([0-2]?[0-9]|3[0-1])\s([0-1]?[0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])\s[0-9]{4}\b"
+# )
```
pederhan, line 17: Can probably remove this if we decide to use the other (more lenient) approach

mreg_cli/utilities/network.py
```diff
@@ -12,57 +12,12 @@
...
-        if last_vlan and network.vlan != last_vlan:
-            return False
-
-        last_vlan = network.vlan
-
-    return True
-
-
 def get_network(ip: str) -> Network | None:
     """Return a network associated with given range or IP."""
```
pederhan, line 22: Is this function still used anywhere? I don't have access to an IDE to check it right now, but my gut feeling says it isn't — and if it is, it's probably a mistake.

pederhan (changes requested):

mreg_cli/utilities/network.py
terjekv: Culled in https://github.com/unioslo/mreg-cli/pull/255/commits/bf4ca3e81b1bca7f4081622d08edb998b35deab2

ci/diff.py
terjekv: Probably, but I'd need to read regular expressions again, so I'm inclined to leave it for now...

terjekv:
Verifying stuff like this is going to take time. Migrating from using indirection (`__`) to direct lookups means that not only the endpoints and results change, but the results go from search/list to direct, so we even change the response from 200 to 404 for what is semantically the same query.

Joy.

````
  [
      {
          "command": "zone create example.org hostmaster@example.org ns1.example.org",
          "command_filter": null,
          "command_filter_negate": false,
          "command_issued": "zone create example.org hostmaster@example.org ns1.example.org  # should require force because ns1 is unknown",
          "ok": [],
          "warning": [
-             "WARNING: : host not found: 'ns1.example.org'",
+             "Host ns1.example.org not found.",
-             "WARNING: : ns1.example.org is not in mreg, must force"
?              -----------
+             "ns1.example.org is not in mreg, must force"
          ],
          "error": [],
          "output": [],
          "api_requests": [
              {
                  "method": "GET",
                  "url": "/api/v1/hosts/ns1.example.org",
                  "data": {},
                  "status": 404,
                  "response": {
                      "detail": "Not found."
                  }
              },
              {
                  "method": "GET",
-                 "url": "/api/v1/hosts/?cnames__name=ns1.example.org",
?                                 -------      ^^^^^^^
+                 "url": "/api/v1/cnames/ns1.example.org",
?                                       ^
                  "data": {},
-                 "status": 200,
?                           ^ ^
+                 "status": 404,
?                           ^ ^
                  "response": {
+                     "detail": "Not found."
-                     "count": 0,
-                     "next": null,
-                     "previous": null,
-                     "results": []
                  }
              }
          ],
          "time": null
      }
  ]
````

**Comment by pederhan** [comment]
```diff
========================================================================
Command: zone delete example.org        -expected, +tested
========================================================================
  [
      {
          "command": "zone delete example.org",
          "command_filter": null,
          "command_filter_negate": false,
          "command_issued": "zone delete example.org",
          "ok": [
-             "OK: : deleted zone example.org"
?              ^^^^^^^
+             "Deleted zone example.org"
?              ^
          ],
          "warning": [],
          "error": [],
          "output": [],
          "api_requests": [
              {
                  "method": "GET",
                  "url": "/api/v1/zones/forward/example.org",
                  "data": {},
                  "status": 200,
                  "response": {
                      "nameservers": [
                          {
                              "name": "ns2.example.org",
                              "ttl": null
                          }
                      ],
                      "updated": true,
                      "primary_ns": "ns2.example.org",
                      "email": "hostperson@example.org",
                      "refresh": 360,
                      "retry": 1800,
                      "expire": 2400,
                      "soa_ttl": 1800,
-                     "default_ttl": 300,
+                     "default_ttl": 43200,
?                                    + +
                      "name": "example.org"
                  }
              }
```

Investigating what caused this...

terjekv:
Nice catch! Who knew tests had a purpose! Then again, 43200? Looks good to me.

mreg_cli/commands/host_submodules/core.py
```diff
@@ -129,6 +128,10 @@ def add(args: argparse.Namespace) -> None:
     host = Host.create(data)
     if not host:
         raise CreateError("Failed to add host.")
+    msg = f"Created host {host.name}"
+    if args.ip:
+        msg += f" with IP {args.ip}"
+    OutputManager().add_ok(msg)
```
pederhan, line 134:
```suggestion
    OutputManager().add_ok(f"Created host {host.name}")
```

pederhan (changes requested):

pederhan:
```diff

========================================================================
Command: network create -network 10.0.1.0/24 -desc "Frozzzen"        -expected, +tested
========================================================================
  [
      {
          "command": "network create -network 10.0.1.0/24 -desc \"Frozzzen\"",
          "command_filter": null,
          "command_filter_negate": false,
          "command_issued": "network create -network 10.0.1.0/24 -desc \"Frozzzen\"        # frozen network",
-         "ok": [
+         "ok": [],
?                ++
-             "OK: : created network 10.0.1.0/24"
+         "warning": [],
+         "error": [
+             "Could not fetch object from location /api/v1/networks/10.0.1.0/24."
          ],
-         "warning": [],
-         "error": [],
          "output": [],
          "api_requests": [
              {
                  "method": "GET",
                  "url": "/api/v1/networks/",
                  "data": {},
                  "status": 200,
                  "response": {
                      "count": 0,
                      "next": null,
                      "previous": null,
                      "results": []
                  }
              },
              {
                  "method": "POST",
                  "url": "/api/v1/networks/",
                  "data": {
                      "network": "10.0.1.0/24",
                      "description": "Frozzzen",
-                     "vlan": null,
-                     "category": null,
-                     "location": null,
                      "frozen": false
                  },
                  "status": 201
+             },
+             {
+                 "method": "GET",
+                 "url": "/api/v1/networks/?id=24",
+                 "data": {},
+                 "status": 200,
+                 "response": {
+                     "count": 0,
+                     "next": null,
+                     "previous": null,
+                     "results": []
+                 }
              }
          ],
          "time": null
      }
  ]
```

The way location for created networks is wrong:

https://github.com/unioslo/mreg-cli/blob/780cec63818008dc64b92e255fedab10f7c4adbf/mreg_cli/api/abstracts.py#L490-L493

This results in an ID of `24` when we create the network `10.0.1.0/24`, since the code splits the resulting location `/api/v1/networks/10.0.2.0/24` on forward slashes and keeps the last segment, and then does a new lookup to `/api/v1/networks/?id=24`, which will fail if no network with the ID `24` exists.

terjekv (dismissed):

pederhan (dismissed):

pederhan (approved):
LGTM 🤠

PR #257, opened 2024-06-04
Copy policies from one host to another.

terjekv:
- This PR allows for copying of all policies (roles) from one host to another via a new command, `policy host_copy`.


pederhan (approved):
LGTM

PR #258, opened 2024-06-05
Fix fetching networks given a location

pederhan:
This PR fixes fetching networks given a location returned by the API. 

In order to make the `Literal` annotation for the `Endpoint.external_id_field()` method work, I had to remove the `@hybridmethod` decorator, but i seemed to be unnecessary anyway? We always call `external_id_field()` on instances of `Endpoint`.

### DIFF
--- a/mreg_cli/api/abstracts.py
+++ b/mreg_cli/api/abstracts.py
@@ -400,6 +400,27 @@ class APIMixin(ABC):
             return None
         return cls(**obj_dict)
 
+    @classmethod
+    def get_by_location(cls, location: str) -> Self | None:
+        """Fetch an object by its location as returned by a `POST` request."""
+        external_id_field = cls.endpoint().external_id_field()
+        if external_id_field == "id":
+            id_ = location.rpartition("/")[-1]
+            if not id_:
+                raise GetError(f"Could not extract ID from location {location}.")
+            try:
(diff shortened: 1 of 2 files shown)


mreg_cli/api/abstracts.py
```diff
@@ -400,6 +391,21 @@ def get_by_query_unique(cls, data: dict[str, str]) -> Self | None:
...
+
+        :param location: The location of the object.
+        :returns: The object if found, None otherwise.
+        """
+        # Some endpoints return invalid location headers,
+        # so we need to look up the object by other means.
+        if location.startswith(Endpoint.Labels):
+            # https://github.com/unioslo/mreg/blob/eed5c154bcc47b1dea474feabad46125ebde0aec/mreg/api/v1/views_labels.py#L30
+            # https://github.com/unioslo/mreg/blob/eed5c154bcc47b1dea474feabad46125ebde0aec/mreg/api/v1/views.py#L187
+            return cls.get_by_field("name", location.rpartition("/")[-1])
```
terjekv, line 406: When https://github.com/unioslo/mreg/issues/532 is fixed, this'll still be correct (as long as our assumptions are the same), but yeah. This is better for now.

terjekv (approved):

PR #259, opened 2024-06-06
Support multiple destinations in host_copy.

terjekv:
- Also reduces output noise.
  - Also handles the situation where a destination already has the role in question.


pederhan (approved):
LGTM

PR #261, opened 2024-06-07
Fix fetching paginated endpoints twice in `get_list_generic`

pederhan:
When writing unit tests for `get_list`, I discovered that we fetch the first page twice when dealing with paginated content that has more than one page of results.

Here's what is currently happening:

1. We fetch, then check if we are dealing with paginated content
2. If we are dealing with paginated content, we fetch the first page again and check for a `next` value.
3. We loop until there are no new `next` values.

This PR eliminates step 2, and enters the while-loop as long as we have a `next` value. If we only have a single page, the loop is never entered.

## Reproducing

By setting the page size to 1 and fetching something known to have more than 1 result, we can see that the first page is fetched twice when adding a print statement to the `get()` function:

```
mreg-test.uio.no> network find -desc *bio*

# Debugger is entered here to set page size
>>> path
'/api/v1/networks/?description__regex=.%2Abio.%2A'
>>> path+="&page_size=1"
>>> path 
'/api/v1/networks/?description__regex=.%2Abio.%2A&page_size=1'

# Application continues executing here
/api/v1/networks/?description__regex=.%2Abio.%2A&page_size=1
/api/v1/networks/?description__regex=.%2Abio.%2A&page_size=1
http://mreg-test.uio.no/api/v1/networks/?description__regex=.%2Abio.%2A&page=2&page_size=1
http://mreg-test.uio.no/api/v1/networks/?description__regex=.%2Abio.%2A&page=3&page_size=1
http://mreg-test.uio.no/api/v1/networks/?description__regex=.%2Abio.%2A&page=4&page_size=1
http://mreg-test.uio.no/api/v1/networks/?description__regex=.%2Abio.%2A&page=5&page_size=1
http://mreg-test.uio.no/api/v1/networks/?description__regex=.%2Abio.%2A&page=6&page_size=1
http://mreg-test.uio.no/api/v1/networks/?description__regex=.%2Abio.%2A&page=7&page_size=1
```

The first page (`/api/v1/networks/?description__regex=.%2Abio.%2A&page_size=1`) is fetched twice in this case.

## Testing

We never caught this in the integration test suite because we don't have enough data for any of the endpoints to return more than one page of results. Pagination unit tests have been added to #251 to ensure pagination works as expected.

### DIFF
--- a/mreg_cli/utilities/api.py
+++ b/mreg_cli/utilities/api.py
@@ -475,23 +475,6 @@ def get_list_generic(
 
     :returns: A list of dictionaries or a dictionary if expect_one_result is True.
     """
-
-    def _check_expect_one_result(
-        ret: list[Json],
-    ) -> Json | list[Json]:
-        if expect_one_result:
-            if len(ret) == 0:
-                return {}
-            if len(ret) != 1:
-                raise MultipleEntititesFound(f"Expected exactly one result, got {len(ret)}.")
-
-            return ret[0]
-
-        return ret
-
-    if params is None:
-        params = {}
-
     response = get(path, params)
 
     # Non-paginated results, return them directly
@@ -511,7 +494,13 @@ def get_list_generic(
             break
         resp = validate_paginated_response(response)
         ret.extend(resp.results)
-    return _check_expect_one_result(ret)
+    if expect_one_result:
+        if len(ret) == 0:
+            return {}
+        if len(ret) != 1:
+            raise MultipleEntititesFound(f"Expected exactly one result, got {len(ret)}.")
+        return ret[0]
+    return ret
 
 
 def get_typed(
--- a/mreg_cli/utilities/api.py
+++ b/mreg_cli/utilities/api.py
@@ -503,25 +503,15 @@ def get_list_generic(
     if limit and resp.count > abs(limit):
         raise TooManyResults(f"Too many hits ({resp.count}), please refine your search criteria.")
 
-    # Short circuit if there are no more pages. This means that there are no more results to
-    # be had so we can return the results we already have.
-    if not resp.next:
-        return _check_expect_one_result(resp.results)
-
     # Iterate over all pages and collect the results
-    ret: list[Json] = []
-    while True:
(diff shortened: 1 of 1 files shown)


mreg_cli/utilities/api.py
```diff
@@ -503,25 +503,15 @@ def _check_expect_one_result(
...
-        if resp is None:
-            return _check_expect_one_result(ret)
-        result = validate_paginated_response(resp)
-
-        ret.extend(result.results)
-
-        if result.next:
-            path = result.next
-        else:
+    ret: list[Json] = resp.results
```
pederhan, line 507: I considered doing a copy here to avoid mutation of the results, but it really has no value. We are returning one large list anyway, so whether we modify the original or a copy doesn't matter to the caller of the function.

terjekv: Makes sense. I think? :)

terjekv (approved):

terjekv:
The only thing I can see here is that we collect all the pages even if we expect one result. We could probably optimise that.

pederhan:
> The only thing I can see here is that we collect all the pages even if we expect one result. We could probably optimise that.

Yes, that is a good point. However, we also need to determine if we have a bad query (more than 1 result when we expect 1). In an extreme scenario where we use a `page_size=1`, how do we determine this if we break after the first page? Look at `count`? Is that reliable? If so, we need to rewrite the result checking. I know this is extreme nit-picking, but we should be careful not to introduce subtle bugs in the name of optimization.

mreg_cli/utilities/api.py
pederhan: Yeah... It seems wrong, but _always_ copying the list when we strictly don't need to is sub-optimal. Then again, computers are fast and network requests are not, so maybe this is a silly concern?

terjekv:
Fix now, optimise later. If people generate a query that expects one result but produces 200 pages, sure, we could have caught that, but so could they. I concur that we go with this clean implementation and (maybe) revisit later.

PR #262, opened 2024-06-10
Fix PyInstaller `host_submodules` imports

pederhan:
This PR fixes the built PyInstaller binaries not including the `host_submodules` commands.

Currently we get this:

```
mreg-test.uio.no> host info
usage: mreg-cli [-h] <command> ...
mreg-cli: error: unrecognized arguments: info
```

This happens because even though PyInstaller follows the `mreg_cli.commands.host_submodules` import, it cannot infer that it should import every module in the `host_submodules` namespace unless we explicitly perform an import of them in `__init__.py`.


pederhan:
> If we first list them manually, we can also do that during the original usage. But, this works. :)

Sure, but then it becomes a bit harder to ensure we actually import all the submodules as opposed to checking `__all__` in a unit test.

PR #266, opened 2024-06-12
Implement logging.

terjekv:
- Also migrates out logging and recording commands to their own command files.

### DIFF
--- a/mreg_cli/config.py
+++ b/mreg_cli/config.py
@@ -22,7 +22,7 @@ import os
 import sys
 from typing import Any, overload
 
-from mreg_cli.types import DefaultType
+from mreg_cli.types import DefaultType, LogLevel
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


mreg_cli/config.py
```diff
@@ -91,10 +86,12 @@ def _load_env_config() -> dict[str, str]:
...
 
     @overload
-    def get(self, key: str) -> str | None: ...
+    def get(self, key: str) -> str | None:
+        ...
 
     @overload
-    def get(self, key: str, default: DefaultType = ...) -> str | DefaultType: ...
+    def get(self, key: str, default: DefaultType = ...) -> str | DefaultType:
+        ...
```
pederhan, line 94: Can you investigate what causes this? My Ruff v0.4.6 formats stub functions on a single line.

terjekv: I did try. And I failed. I think it's VSCode somewhere. `tox -e lint-fix` also doesn't reformat this. We should probably have a pre-commit hook to do this.

pederhan (approved):
LGTM

PR #268, opened 2024-06-19
Colorize diffs with Rich

pederhan:
This PR adds color to the diff produced by `ci/diff.py`:

image

This is a stepping stone to eventually adding interactive diff reviewing similar to `pytest --inline-snapshot=review`:

image

## Compatibility

I stored the terminal output of `diff.py` before and after the code changes, and they produce identical output (except `diff_post` has color when not redirected to a file):

```
# Before:
❯ python diff.py testsuite-result.json new_testsuite_log.json > diff_pre
# After:
❯ python diff.py testsuite-result.json new_testsuite_log.json > diff_post
❯ diff -s diff_pre diff_post
Files diff_pre and diff_post are identical
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,7 @@ jobs:
         run: |
           python -m ensurepip --upgrade
           pip install -r requirements.txt
-          pip install -e .
+          pip install -e .[dev]
       - name: Test and compare api calls
         run: ci/run_testsuite_and_record.sh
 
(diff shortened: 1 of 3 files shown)


oyvindhagberg (approved):
Looks good to me. :+1:

terjekv (approved):
Yep, looks good. Setting TERM=dumb during CI testing may be useful, but this should be fine.

pederhan:
Adding this functionality + reviewing in #270.

PR #269, opened 2024-06-19
Escape error messages

pederhan:
This PR escapes exception messages using prompt_toolkit's own HTML escaping function before passing them to `HTML()`.


pederhan:
Need to iterate on this a bit; we probably want `self.formatted_exception` to be able to provide HTML tags, but we want the content inside those tags to be escaped somehow.

terjekv (approved):
Nice little cleanup. :)

PR #270, opened 2024-06-19
Add interactive diff reviewing

pederhan:
This PR adds interactive reviewing of diff changes.
 
image

Each accepted change is added to `testsuite-result.json`, eliminating the need to rename the test suite files and making it easier to incorporate partial changes.

## Considerations

The PR has a somewhat breaking change in that it writes the new result with all timestamps, IPs, etc. replaced. If we want to write the actual results (i.e without IPs and timestamps replaced) we have to revise this PR considerably.

Running this PR on the existing  `testsuite-results.json` file the first time and selecting "no" to all changes will still modify the file by replacing the aforementioned items.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,6 +11,9 @@ jobs:
         python-version:
           - '3.11'
           - '3.12'
+    env:
+      # Disable colors and formatting in Rich console output
+      TERM: dumb
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -31,7 +34,7 @@ jobs:
         run: |
           python -m ensurepip --upgrade
           pip install -r requirements.txt
-          pip install -e .
+          pip install -e .[dev]
       - name: Test and compare api calls
         run: ci/run_testsuite_and_record.sh
 
--- a/ci/diff.py
+++ b/ci/diff.py
@@ -1,10 +1,22 @@
 from __future__ import annotations
 
+import argparse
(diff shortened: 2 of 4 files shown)


terjekv:
Timestamps are ephemeral, they can happily die in a fire. IPs though... I can see situations where the user expects a consistent behaviour when asking for an IP in a specific way. Ie, if we ask for a specific IP for a host, we should know that tis is the IP we get back. But, well, that is tested by the CLI now, so the output isn't a validator for that per se? Hm.

terjekv (approved):
Concerns about the IPs not withstanding, I really like the idea of this.

terjekv (approved):
LGTM

pederhan:
Getting some different results locally vs in CI:

Local:

```
ERROR: 224 unresolved diff(s) between testsuite-result.json and new_testsuite_log.json.
```

CI:

```
ERROR: 206 unresolved diff(s) between testsuite-result.json and new_testsuite_log.json.
```

Investigating...

terjekv:
> Aha, it's because we now escape error messages, which I implemented in #269...
> FWIW, it renders normally in the console, it just affects the messages we record:

Yeah, that will work after the first time, but it makes the recorded data hard to read. :(

pederhan:
> Yeah, that will work after the first time, but it makes the recorded data hard to read. :(

https://github.com/unioslo/mreg-cli/pull/271

PR #271, opened 2024-06-24
Escape messages only in formatted exceptions

pederhan:
This PR changes the interface for escaping exception messages, allowing us to escape them before passing them to prompt_toolkit, while also allowing us to log the real exception message verbatim.


terjekv (approved):
Yes please, great stuff! Thanks!

PR #274, opened 2024-06-24
Remove `getattr()` usage in `_request_wrapper()`.

pederhan:
This PR adds a `Literal` annotation for the different `requests.Session` methods we invoke in `_request_wrapper`. By declaring the methods we can possibly call, we get better type inference - which in turn removes the need for using `typing.cast`.

This approach is a little more verbose, but makes it much easier to reason with the code as well as improving type safety.

### DIFF
--- a/mreg_cli/utilities/api.py
+++ b/mreg_cli/utilities/api.py
@@ -11,7 +11,7 @@ import logging
 import os
 import re
 import sys
-from typing import Any, Literal, NoReturn, TypeVar, cast, get_origin, overload
+from typing import Any, Literal, NoReturn, TypeVar, get_origin, overload
 from urllib.parse import urljoin
 from uuid import uuid4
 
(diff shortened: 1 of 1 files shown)


mreg_cli/utilities/api.py
```diff
@@ -253,13 +253,23 @@ def _request_wrapper(
...
+    if operation_type == "get":
+        func = session.get
+    elif operation_type == "post":
+        func = session.post
+    elif operation_type == "patch":
+        func = session.patch
+    elif operation_type == "delete":
+        func = session.delete
+    else:
+        raise ValueError(f"Unknown operation type: {operation_type}")
```
pederhan, line 265: This could maybe say "HTTP method" instead of "operation type".

oyvindhagberg (approved):
Looks good!

PR #280, opened 2024-07-02
Fix more minor issues found when looking through test suite results

oyvindhagberg:
- [Add a regex for macaddress to diff.py](https://github.com/unioslo/mreg-cli/commit/ed1212b380be684d1ee582c42f741db36ba35d01).
Macaddresses were caught by the IPv6 regex, which made certain commands in the testsuite look like the wrong type of parameter was supplied.
- [Add OK message to a_add when success](https://github.com/unioslo/mreg-cli/commit/dda10742597b33dfcdaa20b3ed599e478c350eff). Also, avoid refetching the host twice.
- [Remove unnecessary warning from dhcp assoc](https://github.com/unioslo/mreg-cli/commit/e0cd8edd1c8f83a8275c5e1f7af35030a99259ce)
`dhcp assoc <hostname> <ip>` would warn that the hostname isn't a valid ip address. This commit changes the code so the exception isn't raised, but otherwise it works the same. _I am interested in your opinion on whether this is a good solution._


terjekv (approved):
It's an open question where to test the validity of the IP. I'm half tempted to have the test be done by the consumer, but making it transparent-ish like this is also very nice.

PR #282, opened 2024-07-24
Fix `host ttl_remove`

pederhan:
This PR fixes `host ttl_remove` by disabling stripping of `None` values from PATCH data. 

Furthermore, steps have been taken to ensure we never send empty strings as TTL values, and instead preferring to send `None` to these nullable fields.

### DIFF
--- a/mreg_cli/commands/host_submodules/rr.py
+++ b/mreg_cli/commands/host_submodules/rr.py
@@ -863,8 +863,9 @@ def ttl_set(args: argparse.Namespace) -> None:
     valid_ttl = target.valid_ttl_patch_value_with_default(args.ttl)
 
     result = target.patch({"ttl": valid_ttl})
+    new_ttl = result.ttl or args.ttl  # prefer the actual value if it exists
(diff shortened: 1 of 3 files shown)


mreg_cli/commands/host_submodules/rr.py
```diff
@@ -863,8 +863,9 @@ def ttl_set(args: argparse.Namespace) -> None:
     valid_ttl = target.valid_ttl_patch_value_with_default(args.ttl)
 
     result = target.patch({"ttl": valid_ttl})
+    new_ttl = result.ttl or args.ttl  # prefer the actual value if it exists
     if result:
-        OutputManager().add_ok(f"Set TTL for {target} to {args.ttl}.")
+        OutputManager().add_ok(f"Set TTL for {target} to {new_ttl}.")
```
pederhan, line 868: This is just a cosmetic change, wherein we display the *actual* new TTL if the server returns it.

terjekv (approved):

PR #283, opened 2024-07-25
Fix `host set_comment` with empty comment

pederhan:
This PR fixes passing in an empty string as the comment argument to `host set_comment`:

```
user@mreg.example.com> host set_comment myhost ""
ERROR: Patch failure! Tried to set comment to , but server returned None.
```

The error stemmed from the `Host.comment` field converting empty strings to `None`, which then subsequently  caused revalidation to fail because our patch method expected to find an empty string in the `comment` field of the updated model (but got `None`).

This PR removes the nullability of the `comment` field, which matches the [field on the server side](https://github.com/unioslo/mreg/blob/8952b7973c142d90b2c35e37b64c665810237347/mreg/models/host.py#L14).

### DIFF
--- a/mreg_cli/api/models.py
+++ b/mreg_cli/api/models.py
@@ -2493,7 +2493,7 @@ class Host(FrozenModelWithTimestamps, WithTTL, WithHistory, APIMixin):
     bacnetid: int | None = None
     contact: str
     ttl: int | None = None
-    comment: str | None = None
+    comment: str
 
     # Note, we do not use WithZone here as this is optional and we resolve it differently.
     zone: int | None = None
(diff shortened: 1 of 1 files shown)


terjekv (changes requested):
The problem[tm] is that users have made it very clear that they want to be able to "null" this field. I suppose they can set it to `" "`, but I'm not sure that... better? :)

pederhan:
> The problem[tm] is that users have made it very clear that they want to be able to "null" this field. I suppose they can set it to `" "`, but I'm not sure that... better? :)

Sending an empty string (`""`) works and is the only way to remove a comment through the API. Since the field on the model is not nullable, a PATCH request with the following JSON payload is the _only_ way to unset a comment when sending JSON:

```json
{"comment": ""}
```

terjekv:
Ah. Good. Well, eh, good enough? :) Happy for merge.

terjekv (approved):
LGTM

PR #284, opened 2024-07-25
Improve diff numbering + messages

pederhan:
## Numbering

This PR adds a number for each diff for better readability when reviewing them. 

**Number shown before diff**
image

**Number shown when reviewing the change**
image

**Added breathing room between each diff**
image

## Improved messages

We also keep track of the number of resolved and unresolved diffs and print messages and exit accordingly.

**All diffs resolved**
image

**Mix of resolved and unresolved diffs**
image

**No diffs found**
image


terjekv:
Oh yes, great stuff! :)

PR #285, opened 2024-07-25
Use Pydantic v2 style model config in `FrozenModel`

pederhan:
Replaces Pydantic v1-style `class Config` with v2-style `model_config = ConfigDict()` to define a frozen model.


terjekv (approved):
LGTM.

PR #287, opened 2024-07-26
Improve safety of TTL methods for non-nullable fields

pederhan:
This PR makes the methods inherited from `WithTTL` safe to use for models with non-nullable TTL fields (`Zone.default_ttl`& `Zone.soa_ttl`).

Furthermore, a new method `WithTTL.set_ttl()` has been added, which supports both nullable and non-nullable fields.

### DIFF
--- a/mreg_cli/api/models.py
+++ b/mreg_cli/api/models.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import ipaddress
 import re
 from datetime import date, datetime
-from typing import Any, ClassVar, Literal, Self, cast
+from typing import Any, ClassVar, Self, cast
 
 from pydantic import (
     AliasChoices,
@@ -219,8 +219,11 @@ class WithZone(BaseModel, APIMixin):
         return ForwardZone.model_validate(data)
 
 
-class WithTTL(BaseModel):
+class WithTTL(BaseModel, APIMixin):
     """Model for an object that needs to work with TTL values."""
+
+    _ttl_nullable: ClassVar[bool] = True
+    """TTL field(s) of model are nullable."""
 
     @property
     def MAX_TTL(self) -> int:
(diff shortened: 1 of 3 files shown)


mreg_cli/api/models.py
```diff
@@ -384,6 +390,11 @@ class NameServer(FrozenModelWithTimestamps, WithTTL):
     id: int  # noqa: A003
     name: str
 
+    @classmethod
+    def endpoint(cls) -> Endpoint:
+        """Return the endpoint for the class."""
+        return Endpoint.Nameservers
```
pederhan, line 396: Had to be added since `WithTTL` now inherits from `APIMixin`, which requires that an endpoint be defined on its subclasses.

pederhan:
A different approach to determining nullability could be to define a class variable list of non-nullable fields, i.e.:

```py
class WithTTL(BaseModel, APIMixin):
    _non_nullable_ttl_fields: ClassVar[list[str]] = []
    
    def set_ttl(self, ttl: str | int | None, field: str | None = None) -> Self:
        field_name = field or "ttl"
        nullable = field_name in self._non_nullable_ttl_fields
        # ...
    
class Zone(FrozenModelWithTimestamps, WithTTL, APIMixin):
    _non_nullable_ttl_fields: ClassVar[list[str]] = ["soa_ttl", "default_ttl"]
```

It's not necessary right now, but would be the right approach if we had a model that had both nullable and non-nullable TTL fields (WHICH WE SHOULD AVOID AT ALL COSTS).

terjekv:
> if we had a model that had both nullable and non-nullable TTL fields (WHICH WE SHOULD AVOID AT ALL COSTS).

... yeah, let's not do that... Actually, let's not even *think* that.

pederhan:
Merging this, test suite yields identical results (with the exception of non-deterministic addresses causing diffs...).

PR #288, opened 2024-07-29
Replace IDs & IPs in test result URLs

pederhan:
This PR adds regex patterns for replacing IDs and IP adresses in URLs in the test suite results. This adds a band-aid fix for the URLs and IDs currently not being deterministic.

## IDs

`new_testsuite_results.json`:
image

**Result:**
image

Note how only the new query is marked as a difference. The other IDs are replaced by placeholders and thus ignored.

## URL encoded IPs

`new_testsuite_results.json`:
image

**Result:**
image

URL encoded IP addresses are also matched and replaced when comparing.

### DIFF
--- a/ci/diff.py
+++ b/ci/diff.py
@@ -5,6 +5,7 @@ import difflib
 import json
 import re
 import sys
+import urllib.parse
 from enum import StrEnum
 from itertools import zip_longest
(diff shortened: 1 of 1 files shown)


oyvindhagberg (approved):
Very nice. :+1:

PR #289, opened 2024-07-29
A few code changes + changes to testsuite-result

oyvindhagberg:
_Look at the checks! They're all green!_ :green_circle: :green_circle: :green_circle: 
- Changes to `testsuite-result.json` that came from https://github.com/unioslo/mreg-cli/pull/288
- The rest of the changes to `testsuite-result.json`, manually verified
- `a_aaaa.py`: Add a message and a requirement for force when adding an additional ip address to a host.
- `rr.py`: Replace occurrences of str(host.id) with host.id in API requests.
  _Because shouldn't the host id be supplied as a numerical value?_
  This fixes a few differences in the API calls that occur when performing host mx_add and txt_add.

### DIFF
--- a/ci/testsuite-result.json
+++ b/ci/testsuite-result.json
@@ -381,7 +381,7 @@
       },
       {
         "method": "GET",
-        "url": "/api/v1/zones/forward/?id=1",
+        "url": "/api/v1/zones/forward/?id=<ID>",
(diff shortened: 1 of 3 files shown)


pederhan (approved):
LGTM 👏

PR #290, opened 2024-07-30
Write back original lines in diff review

pederhan:
This PR makes diff.py write back the original lines (before placeholders have been inserted) to `testsuite-result.json` when they are accepted in review mode.

The code for loading and processing of the test suite results has been refactored into a class-based approach to reduce the amount of noise when iterating over them in `CommandDiffer.diff_command_results()`.

----

After merging #289, we should do a manual pass where we run the test suite and replace `testsuite-result.json` with the resulting `new_testsuite_log.json` to restore all values replaced by placeholders. This has no effect on the actual tests, but makes it easier for us to search through the log to find specific IPs, MACs, etc.

### DIFF
--- a/ci/diff.py
+++ b/ci/diff.py
@@ -73,36 +73,38 @@ def replace_url_id(match: re.Match[str]) -> str:
     return f"{match.group(1)}{match.group(2)}<ID>{match.group(4)}"
 
 
-def group_objects(json_file_path: str) -> list[dict[str, Any]]:
-    """Group objects in a JSON file by a specific criterion.
+def preprocess_json(s: str) -> str:
+    """Preprocess JSON string for diffing. Replace non-deterministic values with placeholders."""
+    # Replace all URL encoded text in /api/v1/ URLs with unquoted text
+    # This lets us replace it down the line with our normal IPv{4,6} and MAC placeholders
+    # Must be done _before_ all other replacements
+    s = api_v1_pattern.sub(unquote_url, s)
 
-    :param json_file_path: Path to the JSON file.
-    :returns: A list of grouped objects.
-    """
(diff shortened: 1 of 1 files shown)


ci/diff.py
```diff
@@ -155,6 +103,98 @@ def as_string(cls) -> str:
...
+    """Preprocess JSON string for diffing. Replace non-deterministic values with placeholders."""
+    # Replace all URL encoded text in /api/v1/ URLs with unquoted text
+    # This lets us replace it down the line with our normal IPv{4,6} and MAC placeholders
+    # Must be done _before_ all other replacements
+    s = api_v1_pattern.sub(unquote_url, s)
+
+    # Replace all non-deterministic values with placeholders
+    s = timestamp_pattern.sub("<TIME>", s)
+    s = datetime_str_pattern.sub("<TIME>", s)
+    s = serial_pattern.sub("Serial: <NUMBER>", s)
```
terjekv, line 130: Serial here is in context of DNS which is reliant on the date, correct?

terjekv (approved):
LGTM. This Yak must at some point run out of fur... :)

ci/diff.py
pederhan: @oyvindhagberg

terjekv: https://github.com/unioslo/mreg/blob/8952b7973c142d90b2c35e37b64c665810237347/mreg/utils.py#L78-L94 is what I'm thinking of, and I don't think we have any other keys named `Serial` anywhere, so we should be safe from false positives.

oyvindhagberg: Yes, I think you're right.

oyvindhagberg (approved):
Great work!

PR #292, opened 2024-09-02
Use uv in CI

pederhan:
This PR replaces all `pip install` invocations in CI with `uv pip install` using the much faster [uv](https://github.com/astral-sh/uv) Python package manager.


terjekv (approved):
LGTM. There is some irony in redoing all of pythons tooling in Rust because python is too slow for the job. And then writing your software in python. 

My glass house is full of shards and I don't understand why. :D

PR #295, opened 2024-10-02
Add `getattr` default in `APIMixin.refetch`

pederhan:
The `getattr()` call should have a default so we can raise our `InternalError` exception instead of an `AttributeError`.


terjekv (approved):
Yes please, thanks! LGTM.

PR #296, opened 2024-10-02
Omit params when fetching paginated content

pederhan:
Given the path `/api/v1/hostpolicy/roles/` and the params `{"name__regex": "."}`, we send:

```
/api/v1/hostpolicy/roles/?name__regex=.
``` 

And get back this Next URL in the response headers:

```
/api/v1/hostpolicy/roles/?name__regex=.&page=2
``` 

If we pass the original params to `get()` along with the Next URL, we end up requesting the path:

```
/api/v1/hostpolicy/roles/?name__regex=.&page=2&name__regex=.'
```

We should probably just trust that Django gives us a valid Next URL after the first request and omit the original params.

### DIFF
--- a/mreg_cli/utilities/api.py
+++ b/mreg_cli/utilities/api.py
@@ -531,7 +531,7 @@ def get_list_generic(
     # Iterate over all pages and collect the results
     ret: list[Json] = resp.results
     while resp.next:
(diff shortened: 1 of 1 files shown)


terjekv (approved):
LGTM.

PR #300, opened 2024-10-06
Add error msg builder with underlined suggestions

pederhan:
Adds framework for defining error messages with underlined suggestions on how to resolve them. This PR only adds a single builder (for resolving filter errors) to begin with: a builder for resolving regex compilation errors that include a filter (`|`) symbol.

Example:

```
pederhan@mreg.example.com> permission label_add 192.168.0.0/24 oracle-group ^(db|cman)ora.*\.example\.com$ oracle
ERROR: Unable to compile regex 'cman)ora.*\.example\.com$ oracle'
permission label_add 192.168.0.0/24 oracle-group ^(db|cman)ora.*\.example\.com$ oracle
                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                                 └ Consider enclosing this part in quotes.
 ```
 
The interface for determining which builder to use and how to define the message that is raised definitely needs some work. We can consider doing that down the line when/if we expand the error builders, or we can try to work it out in this PR.

## Moving away from InputFailure

In order to format the message as an error with `ERROR:` prepended, `InputFailure` was replaced by `CliError`. We should maybe look into having some sort of `CliError` version of `InputFailure`.

### DIFF
--- a/mreg_cli/errorbuilder.py
+++ b/mreg_cli/errorbuilder.py
@@ -8,6 +8,14 @@ from abc import ABC, abstractmethod
 
 logger = logging.getLogger(__name__)
 
+ExcOrStr = Exception | str
+
+
+# Re-usable function for error builders that don't need a specific implementation
+def default_get_offset(self: "ErrorBuilder") -> tuple[int, int]:  # noqa: ARG001
+    """Get the start and end index of the underlined part of the command."""
+    return -1, -1
+
 
 class ErrorBuilder(ABC):
     """Base class for building error messages with an underline and suggestion."""
@@ -20,10 +28,10 @@ class ErrorBuilder(ABC):
     CHAR_OFFSET = " "
     CHAR_SUGGESTION = "└"
 
-    def __init__(self, command: str, exception: Exception | None = None) -> None:
+    def __init__(self, command: str, exc_or_str: ExcOrStr) -> None:
         """Instantiate the builder with the command that caused the error."""
         self.command = command
-        self.exception = exception
+        self.exc_or_str = exc_or_str
 
     def get_underline(self, start: int, end: int) -> str:
         """Get the underline part of the message."""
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: mreg_cli/outputmanager.py [hunk 1]
```diff
@@ -394,14 +397,14 @@ def get_filter(self, command: str) -> tuple[str, Any, bool]:
...
+                    msg = build_error_message(command_issued)
+                    if msg:
+                        # Custom error messages with suggestions go on their own line
+                        raise CliError(f"{base_msg}\n{msg}") from exc
                     else:
-                        raise InputFailure(
-                            f"Unable to compile regex '{filter_str}': {exc}"
-                        ) from exc
+                        # Otherwise append to the base message
+                        raise CliError(f"{base_msg}: {exc}") from exc
```
pederhan, line 407: This is not ideal in my view. Maybe we should have some sort of `InlineErrorBuilder` that we can fall back on that prints the exception message on the same line if no other builder can be found.

mreg_cli/errorbuilder.py
```diff
@@ -0,0 +1,116 @@
...
+                    )
+                else:
+                    return start, end
+        return -1, -1
+
+    def get_offset(self) -> tuple[int, int]:  # noqa: D102 (missing docstring [inherit it from parent])
+        return self.find_word_with_char_offset(self.command, "|")
+
+
+def get_builder(command: str, exception: Exception | None = None) -> ErrorBuilder | None:
```
pederhan, line 104: This should be refactored either to some sort of mapping lookup, or each Builder could define a `can_build()` (name TBD) class/static method that takes in the command and exception and returns `True`/`False`.

mreg_cli/errorbuilder.py
```diff
@@ -0,0 +1,116 @@
...
+    """Get the appropriate error builder for the given command."""
+    if "|" in command:
+        return FilterErrorBuilder(command, exception)
+    return None
```
pederhan, line 108: This should be refactored either to some sort of mapping lookup, or each Builder could define a `can_build()` (name TBD) class/static method that takes in the command and exception and returns `True`/`False`.

mreg_cli/errorbuilder.py
```diff
@@ -0,0 +1,116 @@
...
+class FilterErrorBuilder(ErrorBuilder):
+    """Error builder for commands with a filter."""
+
+    SUGGESTION = "Consider enclosing this part in quotes."
+
+    @staticmethod
+    def find_word_with_char_offset(command: str, char: str) -> tuple[int, int]:
+        """Find the start and end index of a word containing a specific character.
+
+        Cannot fail; will always return a tuple of two integers.
```
terjekv, line 68: Add a comment about it returning -1, -1 if the target isn't found?

mreg_cli/outputmanager.py
```diff
@@ -394,14 +397,14 @@ def get_filter(self, command: str) -> tuple[str, Any, bool]:
...
                 except re.error as exc:
-                    if "|" in command_issued:
-                        raise InputFailure(
-                            "Command parts that contain a pipe ('|') must be quoted.",
-                        ) from exc
+                    base_msg = f"Unable to compile regex '{filter_str}'"
+                    msg = build_error_message(command_issued)
+                    if msg:
+                        # Custom error messages with suggestions go on their own line
+                        raise CliError(f"{base_msg}\n{msg}") from exc
```
terjekv, line 404: This would probably at some point be factored so one doesn't have to manually remember the newline.

mreg_cli/outputmanager.py
terjekv: Yeah, long term that makes more sense. This is still an improvement!

terjekv (approved):
LGTM

mreg_cli/outputmanager.py
pederhan: Resolved in [a62ad56](https://github.com/unioslo/mreg-cli/pull/300/commits/a62ad561a684dfac634892ad17172ebf2a768c14)

PR #301, opened 2024-10-08
Add defaults for url and domain config options

pederhan:
Sets default values for `--url` and `--domain`. Also comments out `url`, `domain` and `user` from the file `data/mreg-cli.conf` so that local development environments don't load the values from that file when running without `--url`, `--domain`, and `--user`.


terjekv (approved):
For now, we should support stacking config files with inherited defaults in a proper way and distribute the config properly, but yeah, when building a "binary" we're out of luck.

PR #305, opened 2024-10-08
Fix mac assignment in `host add` if provided a network.

### LINKED ISSUE #304: Adding a host to a network with a mac address fails.
```
terjekv-drift@foo> host add -ip 192.168.0.0/23 -contact foo@bar -macaddress aa:bb:cc:dd:ee:ff hostname
'192.168.0.0/23' does not appear to be an IPv4 or IPv6 address
terjekv-drift@foo> host info hostname
Name:         hostname.example.com
Contact:      foo@bar
A_Records     IP               MAC
              192.168.0.50  <not set>
TTL:          (Default)
TXT:          v=spf1 -all
Created:      Tue Oct  8 13:20:50 2024
Updated:      Tue Oct  8 13:20:50 2024
```

terjekv:
- Now properly distinguishes between networks and ips in every case.
  - Also handles the issue where using `host add` with an existing mac address caused a partial creation. The host was created, but the mac wasn't assigned as it was in used and not forced. This now fails early, although there is still a race condition as someone else may have allocated same address between our commands. There is currently no easy fix for this issue.

Fixes #304


pederhan (approved):
LGTM

PR #306, opened 2024-10-10
Build Linux image in redhat/ubi8 container

pederhan:
Also installs git in the container. Had to switch container image because the CentOS repo list doesn't even exist any longer. Let's hope `redhat/ubi8` builds container images that are compatible with RHEL8. You would hope and think so at least.


terjekv (approved):
LGTM

PR #307, opened 2024-10-11
Fix building inside redhat/ubi8 container

pederhan:
This PR makes the following changes to the build and publish workflow:

- Installs git _before_ trying to check out the repo
- Installs `binutils` to get the dependency `objdump`
  - `objdump` was installed by default in the CentOS 8 image, but not in `redhat/ubi8` 🙃
- Fixes the tag pattern to work with tags like `1.0.0a1`, `1.0.0rc1`, etc.

Keeping the commit history so the pain I went through to change 18 lines is recorded for the world.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -3,7 +3,7 @@ name: build mreg-cli
 on:
   push:
     tags:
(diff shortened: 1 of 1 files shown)


terjekv (approved):
LGTM

PR #308, opened 2024-10-11
Refactor `TokenFile`

pederhan:
Simplifies the logic around loading, validating and writing to/from the token file by rewriting the class as a Pydantic model.


terjekv (approved):
LGTM

PR #309, opened 2024-10-16
Fix broken command completion rendering for top-level commands

pederhan:
The number of words in the short help text of top-level commands (`host`, `help`, etc.) would influence the rendering of their CLI completons, leading to `h` displaying `help` and `host` as completions on the same line, making it look like the suggested completion was `help host`.

Very bizarre. To fix this we just disable `display_meta` for top-level command completions. Yeah, not ideal.


terjekv (approved):
Looks terrifying to me, but tests well.

PR #311, opened 2024-10-21
Add prompt string interpolation

pederhan:
Adds string interpolation support for the CLI prompt. In order to test this, the `get_prompt_message()` function has been extracted from `main()`. 

Furthermore, the parsing of command-line config arguments is now moved into the config itself. CLI args are now accessed through `MregCliConfig.get()` instead of via `vars(args)`.


terjekv (approved):
LGTM!

mreg_cli/main.py
```diff
@@ -241,5 +214,46 @@ def get_prompt_message():
...
+            host=host,
+        )
+
+    # Try to get prompt from args -> config -> default
+    if (args_prompt := args_map.get("prompt")) is not None:
+        prompt = str(args_prompt)
+    elif (config_prompt := config.get_prompt()) is not None:
+        prompt = config_prompt
+    else:
+        prompt = DEFAULT_PROMPT
```
pederhan, line 238: This doesn't look _great_, but we have to guard against both `args.prompt` and `config["prompt"]` being None.

README.md
```diff
@@ -38,8 +38,49 @@ Options can be set in ~/.config/mreg-cli.conf. A typical config file looks like
...
+
+The prompt can be specified as a command line argument as well:
+
+```bash
+mreg-cli --prompt "<string>"
+```
+
+#### Disabling the prompt text
+
+The prompt text can be disabled by setting it to an empty string in either the config or as a command line argument:
```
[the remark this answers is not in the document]
pederhan: It means `>`, which is why I specified "prompt _text_". If you have any ideas how to make it clearer, then I'll change it.

mreg_cli/main.py
pederhan: Could call it "prefix" or something.

pederhan:
I _think_ this is ready now? There are some basic unit test for IPv4 and IPv6, but they are not comprehensive at all. Should be fine though.

README.md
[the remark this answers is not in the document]
pederhan: This was changed to "prompt prefix text", which isn't great, but it should remove any ambiguity when combined with the example showing the result.

pederhan:
One last thing we should consider taking a look at is moving the prompt configuration section in the README somewhere else, because it takes up a lot of space very early on for such a trivial feature.

README.md
```diff
@@ -38,10 +38,65 @@ Options can be set in ~/.config/mreg-cli.conf. A typical config file looks like
...
+
+### Prompt
+
+The prompt text can be configured with a custom format string. The available variables are:
+
+- `{user}`: Username of active user
+- `{proto}`: Protocol part of the server URL
+- `{host}`: Name part of the server URL
+- `{port}`: Port part of the server URL (if any)
+- `{domain}`: Domain name. Defaults to `uio.no` if not specified.
```
terjekv, line 56: This was dropped?

terjekv: Ah, right, fun! Thanks!

terjekv (approved):
LGTM!

PR #313, opened 2024-11-07
Reintroduce support for ending network or ip with `/` during `host add`.

terjekv:
- Using / at the end of an IP asks mreg-cli to deduce the netmask.

This reintroduces an old behaviour trait from the previous version of the CLI.

### DIFF
--- a/mreg_cli/commands/host_submodules/core.py
+++ b/mreg_cli/commands/host_submodules/core.py
@@ -123,17 +123,30 @@ def add(args: argparse.Namespace) -> None:
     }
 
     if network_or_ip:
-        network = None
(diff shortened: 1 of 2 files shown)


mreg_cli/commands/host_submodules/core.py
```diff
@@ -88,7 +88,7 @@ def add(args: argparse.Namespace) -> None:
     :param args: argparse.Namespace (name, ip, contact, comment, force, macaddress)
 
     """
-    network_or_ip = args.ip
+    network_or_ip: str = args.ip
```
pederhan, line 91: Explictly declaring this type ensures we don't pass around an `Any` arg anywhere we potentially expect a `NetworkOrIP`.

PR #314, opened 2024-11-08
Fix mac address validation

pederhan:
Replaces regex with [`pydantic_extra_types.mac_address.MacAddress`](https://docs.pydantic.dev/latest/api/pydantic_extra_types_mac_address/).


terjekv (approved):
LGTM, thanks!

PR #315, opened 2024-11-11
Hide traceback on failed login

pederhan:
`LoginFailedError` nows inherits from `CliError`, which causes it to be caught by `Command.parse` and not print a traceback.

Furthermore, `LoginFailedError` and `CliException` are now caught in main(), so that we can use `CliException.print_self` to print a formatted error message when caught.

### DIFF
--- a/mreg_cli/main.py
+++ b/mreg_cli/main.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import argparse
 import getpass
 import logging
-import sys
 
 from prompt_toolkit import HTML
 from prompt_toolkit.shortcuts import CompleteStyle, PromptSession
--- a/mreg_cli/main.py
+++ b/mreg_cli/main.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import argparse
 import getpass
 import logging
+import sys
 
 from prompt_toolkit import HTML
 from prompt_toolkit.shortcuts import CompleteStyle, PromptSession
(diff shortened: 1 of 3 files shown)


mreg_cli/main.py
```diff
@@ -155,10 +155,13 @@ def main():
             str(config.get("url")),
             fail_without_token=args.token_only,
         )
-
     except (EOFError, KeyboardInterrupt, LoginFailedError) as e:
-        print(e)
+        if isinstance(e, LoginFailedError):
```
terjekv, line 159: Is there a reason not to override `__str__` specifically for LoginFailedError to do `print_self()`?

mreg_cli/utilities/api.py
```diff
@@ -161,7 +161,7 @@ def logout() -> None:
 
 def prompt_for_password_and_try_update_token() -> None:
     """Prompt for a password and try to update the token."""
-    password = prompt("You need to re-autenticate\nEnter password: ", is_password=True)
+    password = prompt("You need to re-authenticate\nEnter password: ", is_password=True)
```
terjekv, line 164: *blush* Thanks!

terjekv (approved):
LGTM

mreg_cli/main.py
terjekv: Monday and too much Rust. Ignore me. :)

PR #318, opened 2024-11-12
Make `NETWORK` arg consistent in commands

pederhan:
Since we can specify a network with its network id in `network info` (i.e. 192.168.0.0 can be used to find the network `192.168.0.0/24`), it makes sense to make this behavior consistent across all `network` commands. That way, `network info 192.168.0.0` and `network list_unused_addresses 192.168.0.0` both work by finding the network associated with the address `192.168.0.0`.

A side effect of this new approach, which uses `Network.get_by_any_means_or_raise()`, is that these commands can also find networks using their internal IDs.

### DIFF
--- a/mreg_cli/commands/network.py
+++ b/mreg_cli/commands/network.py
@@ -238,7 +238,7 @@ def list_unused_addresses(args: argparse.Namespace) -> None:
 
     :param args: argparse.Namespace (network)
     """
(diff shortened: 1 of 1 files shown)


terjekv (approved):
LGTM. What can possibly go wrong? 😄

PR #319, opened 2024-11-15
Add consistent constructors for `IPAddressField`, `MACAddressField` & `NetworkOrIP`

pederhan:
Adds the method `validate()` for models `IPAddressField`, `MACAddressField` & `NetworkOrIP`, which takes care of validation and exception handling when instantiating these types. This removes a lot of boilerplate code used for validating IPs and MACs.

Furthermore, a new method `NetworkOrIP.parse()` has been added, which allows for validation of an argument as a specific type of IP address or network. I.e. we can instantiate `ipaddress.IPv{4,6}{Network,Address}` using a single method

```py
# Version agnostic
ip = NetworkOrIP.parse("192.168.0.1", mode="ip")
net = NetworkOrIP.parse("192.168.0.0/24", mode="network")

# v4
ipv4 = NetworkOrIP.parse("192.168.0.1", mode="ipv4")
netv4 = NetworkOrIP.parse("192.168.0.0/24", mode="networkv4")

# v6
ipv6 = NetworkOrIP.parse("2001:db8::1", mode="ipv6")
netv6 = NetworkOrIP.parse("2001:db8::/64", mode="networkv6")
```

This takes care of exception handling for us:

```py
ip = NetworkOrIP.parse("192.168.0.0/24", mode="ip")
```

```
Traceback (most recent call last):
  File "/Users/pederhan/.vscode/extensions/ms-python.python-2024.21.2024111202-darwin-arm64/python_files/python_server.py", line 130, in exec_user_input
    retval = callable_(user_input, user_globals)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 1, in <module>
  File "/Users/pederhan/repos/mreg-cli/mreg_cli/api/models.py", line 127, in parse
    return func(ipnet)
           ^^^^^^^^^^^
  File "/Users/pederhan/repos/mreg-cli/mreg_cli/api/models.py", line 174, in as_ip
    raise InvalidIPAddress(f"{self.ip_or_network} is not an IP address.")
mreg_cli.exceptions.InvalidIPAddress: 192.168.0.0/24 is not an IP address.

```

One can also omit the `mode=` argument entirely to construct a `IP_AddressT | IP_NetworkT` directly without further type narrowing:

```py
ip = NetworkOrIP.parse("192.168.0.1")
net = NetworkOrIP.parse("192.168.0.0/24")
```

## Discussion

Since the various `validate()` and `parse()` methods raise `InputFailure` whenever they fail, we end up adding a warning every time validation fails, even if we never re-raise the error. This is a problem, because we always try to construct a MAC address from the identifier when looking up hosts in `Host.get_by_any_means` when we dont have a `HostT` hostname, id or IP.

We can either live with an increased number of warnings in the JSON records, or we will have to manually construct a `MacAddressField` in `Host.get_by_any_means` instead of using the common `validate` method, which I'm sure will come back to bite us eventually.

EDIT: this was solved by adding `MACAddressField.validate_naive`, which raises `ValueError` when failing instead of `InputError`, so we can perform these [EAFP](https://docs.python.org/3.12/glossary.html#term-EAFP) checks without logging them:

https://github.com/unioslo/mreg-cli/blob/c8e3989aa9f0eb5de887e6a558f2e763b61361c1/mreg_cli/api/fields.py#L29-L39

It's a gross hack, but it works if silencing failed mac address validation is the goal.

### DIFF
--- a/mreg_cli/api/models.py
+++ b/mreg_cli/api/models.py
@@ -2697,7 +2697,7 @@ class Host(FrozenModelWithTimestamps, WithTTL, WithHistory, APIMixin):
             try:
                 mac = MACAddressField.validate(identifier)
                 return Host.get_by_field("ipaddresses__macaddress", mac.address)
-            except ValueError:
+            except InputFailure:
                 pass
 
             # Let us try to find the host by name...
--- a/mreg_cli/api/models.py
+++ b/mreg_cli/api/models.py
@@ -5,7 +5,8 @@ from __future__ import annotations
 import ipaddress
 import re
 from datetime import date, datetime
-from typing import Any, ClassVar, Self, cast
+from enum import StrEnum
+from typing import Any, ClassVar, Literal, Self, cast, overload
 
 from pydantic import (
     AliasChoices,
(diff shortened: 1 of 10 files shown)


mreg_cli/api/models.py
```diff
@@ -55,13 +55,77 @@
...
+        """Parse a value as an IP address or network.
+
+        Optionally specify the mode to validate the input as.
+
+        :param value:The value to parse.
+        :param mode: The mode to validate the input as.
+        :returns: The parsed value as an IP address or network.
+        """
+        ipnet = cls.validate(value)
+        funcmap: dict[IPNetMode, Callable[..., IP_AddressT | IP_NetworkT]] = {
```
terjekv, line 118: Okay, this is neat. A very nice solution!

mreg_cli/api/models.py
```diff
@@ -55,13 +55,77 @@
...
 
 
 class NetworkOrIP(BaseModel):
     """A model for either a network or an IP address."""
 
-    ip_or_network: str | IP_AddressT | IP_NetworkT
+    ip_or_network: IP_AddressT | IP_NetworkT
+
+    @classmethod
+    def validate(cls, value: str | IP_AddressT | IP_NetworkT | Self) -> Self:
```
terjekv, line 67: As this returns an instance, is this `validate()` or `new()`? It feels a little weird to call it for instantiation.

terjekv (approved):
Very nice cleanup! Thank!

mreg_cli/api/models.py
pederhan: If this was Go, I would definitely have named it `New()`, however, there isn't an agreed-upon naming scheme for alternate constructors in Python as far as I'm aware, so that passed me by. I decided to use the now-deprecated Pydantic v1 method name `validate()` which takes in any value, validates it, and returns an instance of the model.

We can change the name if you want. I don't mind either way. I just needed something better than `from_string`, since we can validate a lot more than just strings.

terjekv: Let's call it a nod to Pydantic v1 and keep it. :)

PR #320, opened 2024-11-15
Improve `NameList` robustness and flexibility

pederhan:
This PR makes the custom `NameList` type less brittle by not breaking when receiving values without a `"name"` key. The validator instead opts to filter out these dicts and log an error when it happens. 

Tests have been added for a small variety of different inputs to the field.


terjekv (approved):
LGTM

PR #321, opened 2024-11-16
Add uv lock file

pederhan:
Just like unioslo/mreg#553, this PR moves all dependency to uv. This simplifies managing dependency upgrades and ensures consistent environments between developers and in CI.

This removes the need to manage requirements*.txt files.


terjekv (approved):
LGTM

PR #324, opened 2024-11-17
Update changelog

pederhan:
Updates changelog. Note that the last point in **Fixed** regarding network addresses assigned to hosts is not correct (yet). After I wrote that, I tested it and found it to be incorrect.


terjekv (approved):
LGTM

PR #326, opened 2024-11-17
Fix assigning network addresses to hosts

### LINKED ISSUE #325: Hosts can be assigned network addresses
It is possible to assign network addresses to hosts when creating them. `list_excluded_addresses` states that the network address is reserved, yet it is possible to assign it to a host.

pederhan:
Fixes #325

*However*, I feel like this is the wrong way to go about it. It should not be the CLI's job to babysit the details of user input to this level of granularity. 

We can implement this PR as a stop-gap until it's fixed on the server side, but it's not great having to go back and undo these changes once it's fixed on the server.


terjekv (approved):
LSTM. (looks sad to me). But needed for now.

pederhan:
I realized the exact same bug applied to broadcast addresses, so instead of adding a bunch of individual checks, we now simply check if the address is reserved. The list of reserved addresses includes both network and broadcast addresses.

terjekv:
> I realized the exact same bug applied to broadcast addresses, so instead of adding a bunch of individual checks, we now simply check if the address is reserved. The list of reserved addresses includes both network and broadcast addresses.

Hm. There are some force/admin rules here...

`is_permitted_to_use_ipaddress_or_raise` from
https://github.com/unioslo/mreg/pull/554/files#diff-0895a17194174fc6c87b8a5dcc2bd73915273c8c1b018f950ea6169032c61ab9 is worth a read, that maps to the current tests done by the server.

pederhan:
Ok, then we'll go back to network and broadcast, but trust the server to handle other reserved addresses.

PR #327, opened 2024-11-17
Remove `Network.str_to_network`

pederhan:
Replaces all `Network.str_to_network()` calls with `NetworkOrIP.parse(..., mode="network")`, since it fulfils the same job. That way we get consistent error messages for all such operations, and we also remove some code in the process, which is always good.


terjekv (approved):
Good cleanup, thanks!

PR #329, opened 2024-11-18
Add version number in user agent header

pederhan:
Adds mreg-cli version number in the user agent header for each request. 

So instead of `mreg-cli` we now send `mreg-cli-x.y.z` in the user agent field.


terjekv (approved):
LGTM

PR #330, opened 2024-11-19
Freeze lock file in CI

pederhan:
This PR freezes the lock file in CI to avoid updating it when running `uv sync` .


terjekv (approved):
That poor Yak.

PR #333, opened 2024-11-20
Require successful PyPI publishing before creating release

pederhan:
Since we include PyPI installation instructions in the release notes, we should make sure the PyPI release actually, you know, exists.


terjekv (approved):
:D

PR #335, opened 2024-11-20
Re-add command completion descriptions

pederhan:
This reverts the "fix" applied in #309 . I have no idea how to make the `host` and `help` completions render vertically, when `permission` and `policy` already do so by default.


terjekv (approved):
LGTM

Issue #323, opened 2024-11-17
Generate coverage report from unit and integration tests


pederhan:
It would be useful to generate a coverage report from a combination of unit tests and the integration test suite. That would give us an overview of commands that are in need of better testing. 

Seeing as the breakage fixed in #322 slipped under the radar, it means that command is neither tested by unit tests or integration tests. Having proper coverage reporting would have warned us of that command (and others) not being tested.

PR #338, opened 2024-11-21
Fix `host info` for host with multiple IPs on same MAC

### LINKED ISSUE #337: Looking up a host by MAC address should check if multiple MACs resolve to the same host.
If a host has multiple IPs and they are associated to the same MAC we can get:
```
foo@bar> host info 50:9a:4c:46:1e:6b
Expected exactly one result, got 2.
```

pederhan:
Fixes #337 

We should probably have a test in the test suite for this.

### DIFF
--- a/ci/testsuite-result.json
+++ b/ci/testsuite-result.json
@@ -25049,7 +25049,7 @@
     "api_requests": [
       {
         "method": "GET",
-        "url": "/api/v1/hosts/?ipaddresses__ipaddress=10.0.0.20&ordering=name",
+        "url": "/api/v1/hosts/?ipaddresses__ipaddress=10.0.0.20",
         "data": {},
         "status": 200,
         "response": {
--- a/ci/testsuite-result.json
+++ b/ci/testsuite-result.json
@@ -20023,6 +20023,127 @@
     "time": null
   },
   {
+    "command": "host info foo",
+    "command_filter": null,
+    "command_filter_negate": false,
+    "command_issued": "host info foo # should work and show both IPs",
+    "ok": [],
+    "warning": [],
+    "error": [],
+    "output": [
+      "Name:         foo.example.org",
+      "Contact:      ",
+      "A_Records     IP            MAC",
+      "              10.0.0.5      aa:bb:cc:dd:ee:ff",
+      "AAAA_Records  IP            MAC",
+      "              2001:db8::5   <not set>",
+      "TTL:          (Default)",
+      "TXT:          v=spf1 -all",
+      "Created:      Thu Nov 21 10:34:55 2024",
+      "Updated:      Thu Nov 21 10:34:55 2024"
+    ],
+    "api_requests": [
+      {
+        "method": "GET",
+        "url": "/api/v1/hosts/foo.example.org",
+        "data": {},
+        "status": 200,
+        "response": {
+          "ipaddresses": [
+            {
+              "macaddress": "aa:bb:cc:dd:ee:ff",
+              "ipaddress": "10.0.0.5",
+              "host": 13
+            },
+            {
+              "macaddress": "",
+              "ipaddress": "2001:db8::5",
+              "host": 13
+            }
+          ],
+          "cnames": [],
+          "mxs": [],
+          "txts": [
+            {
+              "txt": "v=spf1 -all",
+              "host": 13
+            }
+          ],
+          "ptr_overrides": [],
+          "hinfo": null,
+          "loc": null,
+          "bacnetid": null,
+          "name": "foo.example.org",
+          "contact": "",
+          "ttl": null,
+          "comment": "",
+          "zone": 1
+        }
+      },
+      {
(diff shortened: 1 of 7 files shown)


mreg_cli/utilities/api.py
```diff
@@ -540,7 +540,7 @@ def get_list_generic(
     if expect_one_result:
         if len(ret) == 0:
             return {}
-        if len(ret) != 1:
+        if len(ret) > 1 and not all(ret[0] == x for x in ret):
             raise MultipleEntititesFound(f"Expected exactly one result, got {len(ret)}.")
```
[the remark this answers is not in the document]
pederhan: Yeah, can do.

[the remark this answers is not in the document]
pederhan: Well, we don't know that they are _all_ distinct from each other, but I can't think of a good message that conveys the ambiguity involved in interpreting the results.

mreg_cli/utilities/api.py
```diff
@@ -540,7 +540,7 @@ def get_list_generic(
     if expect_one_result:
         if len(ret) == 0:
             return {}
-        if len(ret) != 1:
+        if len(ret) > 1 and not all(ret[0] == x for x in ret):
```
oyvindkolbu, line 543: Suggest changing `not all` to `any` and thus the `==` to `!=`.

mreg_cli/utilities/api.py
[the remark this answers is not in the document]
terjekv: We could theoretically check for the number of truly distinct matches, but I'm not sure that'd make things any clearer. :)

mreg_cli/utilities/api.py
pederhan: Yes, that short circuits faster, you are right.

ci/testsuite-result.json
```diff
@@ -20022,6 +20022,127 @@
...
+    "ok": [],
+    "warning": [],
+    "error": [],
+    "output": [
+      "Name:         foo.example.org",
+      "Contact:      ",
+      "A_Records     IP            MAC",
+      "              10.0.0.5      aa:bb:cc:dd:ee:ff",
+      "AAAA_Records  IP            MAC",
+      "              2001:db8::5   <not set>",
```
pederhan, line 20039: Need to fix MAC association here

terjekv: Do we? I mean, it's implied due to the way things work at UiO, but the association is not explicitly set and I don't think that was the intent.

pederhan: Ok, let's leave it as it is.

mreg_cli/api/models.py
```diff
@@ -2701,7 +2701,7 @@ def get_by_any_means(
                     return hosts[0]
 
                 if len(hosts) > 1:
-                    raise MultipleEntititesFound(
+                    raise MultipleEntitiesFound(
```
pederhan, line 2704: Could we hit the same error here?

mreg_cli/api/models.py
```diff
@@ -2701,7 +2701,7 @@ def get_by_any_means(
...
                         f"Multiple hosts found with IP address or PTR {identifier}."
```
pederhan, line 2705: Could we hit the same error here?

terjekv: Yes, indeed.

mreg_cli/api/models.py
```diff
@@ -2701,7 +2701,7 @@ def get_by_any_means(
                     return hosts[0]
 
                 if len(hosts) > 1:
```
pederhan, line 2703: We could do this:

```suggestion
                if len(hosts) > 1 and any(hosts[0].id != host.id for host in hosts):
```

But should we consider refactoring the method used to fetch hosts to `Host.get_list_by_query_unique`, and then instead catch `MultipleEntitiesFound` and raise the more granular `MultipleEntitiesFound` we already have? Something like:


```py
            try:
                ptr = False
                ipaddress.ip_address(identifier)

                host = Host.get_by_query_unique({"ipaddresses__ipaddress": identifier})
                if not host:
                    host = Host.get_by_query_unique({"ptr_overrides__ipaddress": identifier})
                    ptr = True

                if host:
                    if ptr and inform_as_ptr:
                        OutputManager().add_line(f"{identifier} is a PTR override for {host.name}")
                    return host
            except MultipleEntitiesFound as e:
                raise MultipleEntitiesFound(
                    f"Multiple hosts found with IP address or PTR {identifier}."
                ) from e
            except ValueError: # invalid IP (should use NetworkOrIP.parse...)
                pass
```

pederhan: I have pushed a commit which implements this. I do think we need more comprehensive tests for a broader range of host/ip/mac/network permutations though. I'll look into it once we have implemented #323.

terjekv (approved):
LGTM

PR #342, opened 2024-11-22
Support for help whoami and help versions.

terjekv:
Lots of the useful data from the server relies on https://github.com/unioslo/mreg/pull/562, but the client gracefully handles missing endpoints.


pederhan (approved):
LGTM!

PR #343, opened 2024-11-22
Fix underline in errors for commands with regex as final arg

pederhan:
The `>=` check for the `end` offset was wrong. It should have been `>`. This caused the underline for those commands to be considered invalid, and defaulted to `(-1, -1)`, which meant no underline was rendered.

Also rewrote tests as snapshot tests so we don't have to write the expected results by hand.


terjekv (approved):
LGTM

PR #344, opened 2024-11-23
Fix parsing of http://localhost:8000 for the prompt.

terjekv:
- Simplifies the regex.
  - Cleanly allows empty domain.
  - Added tests.


pederhan (approved):
Good improvement!

PR #345, opened 2024-11-23
Relax markdownlint settings from the default and add linting action

terjekv:
Adds rules for markdownlint to the repo and a step to lint against those rules during testing.


pederhan (approved):
LGTM!

PR #346, opened 2024-11-27
Format `help {whois,whoami}` permissions as table

pederhan:
Also adds a new option `-django` to optionally include internal django roles.

### DIFF
--- a/mreg_cli/config.py
+++ b/mreg_cli/config.py
@@ -31,8 +31,6 @@ from mreg_cli.types import DefaultType, LogLevel
 logger = logging.getLogger(__name__)
 
 
-p = platformdirs.PlatformDirs("mreg-cli", "UiO")
-
 # Config file locations.
 # This needs migration to platformdirs... Without breaking historical usage.
 DEFAULT_CONFIG_PATH = tuple(


mreg_cli/config.py
```diff
@@ -241,6 +244,10 @@ def get_category_tags(self) -> list[str]:
         """Get the category tags from the application."""
         return self.get("category_tags", "").split(",")
 
+    def get_user(self) -> str | None:
+        """Get the user from the application."""
+        return self.get("user")
+
```
pederhan, line 250: Future PR should replace calls to `MregCliConfig.get("user")` with this method.

mreg_cli/config.py
```diff
@@ -30,6 +30,9 @@
 
 logger = logging.getLogger(__name__)
 
+
+p = platformdirs.PlatformDirs("mreg-cli", "UiO")
```
terjekv, line 34: Where is this used?

terjekv (approved):

mreg_cli/outputmanager.py
```diff
@@ -405,9 +404,9 @@ def get_filter(self, command: str) -> tuple[str, Any, bool]:
 
     def add_formatted_table(
         self,
-        headers: Sequence[str],
-        keys: Sequence[str],
-        data: Sequence[dict[str, Any] | BaseModel],
+        headers: Iterable[str],
+        keys: Iterable[str],
+        data: Iterable[dict[str, Any] | BaseModel],
```
pederhan, line 409: Iterable in case we have some sort of comprehension or generator that we pass in. We don't need access to indexing. We can remove this, I don't feel too strongly about it, but one could imagine passing in a generator expression or similar to `output_multiple`, which does not fulfill the `Sequence` interface.

mreg_cli/config.py
pederhan, line 34: Will revert, dirty repo.

pederhan: I posted a comment about this at the same time: https://github.com/unioslo/mreg-cli/pull/346#discussion_r1860533626

mreg_cli/outputmanager.py
terjekv: Iterable is a good solution. :)

terjekv (approved):
LGTM!

PR #347, opened 2024-11-27
Improve `FilterErrorBuilder.can_build` heuristics

pederhan:
Declares a `DEFAULT_OFFSET` value we can compare against. Decorates `find_word_with_char_offset` with `@cache` to let us calculate the offsets, check offsets, then call the builder again, without having to recalculate the offsets.

### DIFF
--- a/mreg_cli/errorbuilder.py
+++ b/mreg_cli/errorbuilder.py
@@ -2,20 +2,17 @@
 
 from __future__ import annotations
 
-from functools import cache
 import logging
 import shlex
 from abc import ABC, abstractmethod
+from functools import cache
 
 logger = logging.getLogger(__name__)
 
 ExcOrStr = Exception | str
 
 
-# Re-usable function for error builders that don't need a specific implementation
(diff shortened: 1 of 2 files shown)


mreg_cli/errorbuilder.py
```diff
@@ -107,12 +108,16 @@ def can_build(self) -> bool:  # noqa: D102
         return True
 
 
+DEFAULT_OFFSET = (-1, -1)
```
terjekv, line 111: Nitpicky, but is this really a "default_offset" or a "word_not_found_result"?

terjekv (approved):
LGTM

mreg_cli/errorbuilder.py
pederhan: You are right that it is more of a sentinel value - albeit one that still contains valid indices. I have refactored it to be more clear now.

terjekv: The refactored code looks good! Thanks!

PR #349, opened 2024-11-29
Fix IPv6 host lookup raising `ValidationError` when not found

### LINKED ISSUE #348: `host info` with IPv6 argument raises `ValidationError` when host is not found
Testing with some random IPv6 address:

```
super@mreg-dev.uio.no> host info 03ec:abd3:d906:f309:d0fe:9dcc:4ff0:7da2
1 validation error for HostT
hostname
  Value error, Invalid input for hostname: 03ec:abd3:d906:f309:d0fe:9dcc:4ff0:7da2 [type=value_error, input_value='03ec:abd3:d906:f309:d0fe:9dcc:4ff0:7da2', input_type=str]
    For further information visit https://errors.pydantic.dev/2.9/v/value_error
```

Which happens, because we try to fall back on interpreting the argument as a hostname:

https://github.com/unioslo/mreg-cli/blob/14e54fc6f508b358a29cd686a4e76e2164745501/mreg_cli/api/models.py#L2705-L2706

But interpreting an IPv6 address as a host name causes a ValidationError due to our custom validator:

https://github.com/unioslo/mreg-cli/blob/14e54fc6f508b358a29cd686a4e76e2164745501/mreg_cli/api/models.py#L226-L227

pederhan:
This PR fixes #348 by treating a failed lookup in `Host.get_by_any_means_or_raise` for IP/MAC with a _valid_ IP/MAC as "host not found", rather than proceeding by interpreting the argument as a hostname. This _does_ break support for hosts with valid IPs/MACs used as hostnames, but that should probably not be supported anyway (?)

## Problems

This is a draft PR, because it causes a bunch of new warnings to be added to the test suite results. Every time we check if a IP or MAC is valid, and it's not, we add a warning to the log. This happens because our helper functions raise `CliWarning` or `CliError`, and it's not possible to use them to check if something is valid, then catch the exception silently. Those exception types _always_ log when they are instantiated.

If this is fine, then sure, we can just merge the PR with the extra warnings in the result log, but if not, we need to consider a new scheme for logging exceptions.

I think automatic recording of every exception can cause a lot of test suite result churn, and we should take a look at it.

### DIFF
--- a/mreg_cli/api/models.py
+++ b/mreg_cli/api/models.py
@@ -10,6 +10,7 @@ from functools import cached_property
 from typing import Any, Callable, ClassVar, Iterable, List, Literal, Self, cast, overload
 
 from pydantic import AliasChoices, BaseModel, ConfigDict, Field, computed_field, field_validator
+from pydantic_extra_types.mac_address import MacAddress
 from typing_extensions import Unpack
 
 from mreg_cli.api.abstracts import APIMixin, FrozenModel, FrozenModelWithTimestamps
@@ -100,7 +101,11 @@ class NetworkOrIP(BaseModel):
 
     @overload
     @classmethod
-    def parse(cls, value: Any, mode: IPNetMode) -> IP_AddressT | IP_NetworkT: ...
+    def parse(cls, value: Any, mode: Literal["networkv4"]) -> ipaddress.IPv4Network: ...
+
+    @overload
+    @classmethod
+    def parse(cls, value: Any, mode: Literal["networkv6"]) -> ipaddress.IPv6Network: ...
 
     @classmethod
     def parse(cls, value: Any, mode: IPNetMode | None = None) -> IP_AddressT | IP_NetworkT:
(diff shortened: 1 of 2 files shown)


mreg_cli/api/models.py
```diff
@@ -126,6 +131,38 @@ def parse(cls, value: Any, mode: IPNetMode | None = None) -> IP_AddressT | IP_Ne
...
+        mode = "ipv4" if version == 4 else "ipv6" if version == 6 else "ip"
+        try:
+            return cls.parse(ip, mode=mode)
+        except ValueError:
+            return None
+
+    @classmethod
+    def parse_network_optional(
+        cls, ip: str, version: Literal[4, 6] | None = None
+    ) -> IP_NetworkT | None:
```
pederhan, line 153: Added to match `parse_ip_optional`, but it's unused (as of now).  Adding unused stuff sucks, but if we want to use `parse_ip_optional`, it's likely we want to use `parse_network_optional` somewhere too.

mreg_cli/api/models.py
```diff
@@ -126,6 +131,38 @@ def parse(cls, value: Any, mode: IPNetMode | None = None) -> IP_AddressT | IP_Ne
             return func(ipnet)
         return ipnet.ip_or_network
 
+    @classmethod
+    def parse_ip_optional(
+        cls, ip: str, version: Literal[4, 6] | None = None
+    ) -> IP_AddressT | None:
```
pederhan, line 137: This really shouldn't take a `version` argument, but should instead have a bunch of overloads like `parse` has. However, I got lazy. Will go back and fix.

pederhan: Also, this serves a lot of the same purpose as `IPAddressField.is_valid_ip`, except it returns the IP address as well. They are used in much the same way, except in some places we want to keep the resulting IP (if it exists). 

We should look into unifying this into a single method. It wouldn't even break the current conditionals i.e.:

```
- if IPAddressField.is_valid_ip(ip):
+ if NetworkOrIP.parse_ip_optional(ip):
```

terjekv: Are our semantics closer to `parse` vs `parse_or_raise`?

pederhan: Yes, we should rename it and refactor code that uses what is currently called `parse`.

pederhan:
I think we'll fix the `parse`/`parse_or_raise` semantics in a future PR. I am working on it, but it touches on a lot of commands, so it's best to not mix those changes with this PR.

pederhan:
```
super@mreg-dev.uio.no> host info 2176:f6fd:bd73:3e62:fcd6:7bda:8335:f22b
Host with IP address 2176:f6fd:bd73:3e62:fcd6:7bda:8335:f22b not found.
```

terjekv (approved):

PR #350, opened 2024-11-30
Only log exceptions when caught

pederhan:
This PR refactors logging of exceptions by adding the new methods `log()` and `print_and_log()` to `CliException`, and removing the logging + `OutputManager.add_{warning,error}` calls in the constructor. 

All logging is moved to the new `log()` method, which can be overridden in subclasses to control how a given exception type is logged. The new `print_and_log()` method simply calls `log()`and `print_self()`.

### DIFF
--- a/ci/testsuite-result.json
+++ b/ci/testsuite-result.json
@@ -6,7 +6,6 @@
     "command_issued": "zone create example.org hostmaster@example.org ns1.example.org  # should require force because ns1 is unknown",
     "ok": [],
     "warning": [
-      "Host ns1.example.org not found.",
       "ns1.example.org is not in mreg, must force"
     ],
     "error": [],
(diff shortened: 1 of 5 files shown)


ci/testsuite-result.json
```diff
@@ -210,10 +207,9 @@
     "command": "zone set_ns example.org ns2.example.org",
     "command_filter": null,
     "command_filter_negate": false,
-    "command_issued": "zone set_ns example.org ns2.example.org # requires force because ns2 is unknown",
+    "command_issued": "zone set_ns example.org ns2.example.org  #  requires force because ns2 is unknown",
```
pederhan, line 210: A little bit unsure why this changed. I guess we ignore whitespace somewhere, and we never had a diff until now that caused this command to be considered "different".

terjekv (approved):
This looks like a good approach!

PR #352, opened 2024-12-01
Refactor `parse()` methods to conform with naming conventions

pederhan:
Renames the following methods:

-  `MACAddressField.parse()` -> `MACAddressField.parse_or_raise()`
-  `MACAddressField.parse_optional()` -> `MACAddressField.parse()`
-  `NetworkOrIP.parse()` -> `NetworkOrIP.parse_or_raise()`
-  `NetworkOrIP.parse_optional()` -> `NetworkOrIP.parse()`

Also fixes a bug where `NetworkOrIP.validate()` could raise a `pydantic.ValidationError` on failed validation instead of `InputFailure` because we caught our own `ValidationError` type instead of Pydantic's.

### DIFF
--- a/tests/api/test_models.py
+++ b/tests/api/test_models.py
@@ -86,7 +86,7 @@ def test_network_or_ip_parse(inp: str, mode: IPNetMode, expect: Any) -> None:
 
 
 @pytest.mark.parametrize(
-    "inp, expect_type",
+    "inp, expect_type_call",
     [
         ("192.168.0.1", NetworkOrIP.is_ipv4),
         ("192.168.0.0/24", NetworkOrIP.is_ipv4_network),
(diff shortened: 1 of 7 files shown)


tests/api/test_models.py
```diff
@@ -80,10 +81,33 @@
 )
 def test_network_or_ip_parse(inp: str, mode: IPNetMode, expect: Any) -> None:
     """Test the network or IP address from string."""
-    res = NetworkOrIP.parse(inp, mode)
+    res = NetworkOrIP.parse_or_raise(inp, mode)
     assert res == expect
 
 
+@pytest.mark.parametrize(
+    "inp, expect_type",
```
terjekv, line 89: No biggie, but I'd prefer this to be `expect_type_call` to point out that this parameters ia a callable and not a type.

terjekv (approved):
LGTM. Long term we might want to consider making the acceptable modes for `parse` and `parse_or_raise` into an enum instead of typed literals, but that's for the future. :)

PR #353, opened 2024-12-03
Containerize setup for testing

oyvindhagberg:
This PR moves UV and Python setup/installation from the workflow and into a Dockerfile. A new test script is added that builds a container for mreg-cli with all the dependencies.
The purpose is to make it easier to run the testsuite. Now you just have to clone the repo and run the script, as long as you have docker or podman installed. This is great for other workflows that need to do this when building mreg.

This PR also adds Python version 3.13 to the matrix for unit- and integration tests. Tried it with tox, but it didn't work.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,20 +14,15 @@ jobs:
         python-version:
           - '3.11'
           - '3.12'
+          - '3.13'
     env:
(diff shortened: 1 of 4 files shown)


terjekv (approved):
LGTM :)

pederhan (approved):
LGTM

PR #355, opened 2024-12-03
Use original test suite result when writing diffs

pederhan:
I guess when I originally wrote the new diff.py I wanted to be on the safe side and use the new result rather than the original one, but that leads to a lot of unnecessary changes to the test suite results file.

This PR selects the old result when writing the diff file, so that we only modify the lines pertaining to the commands that have differences, instead of updating a bunch of lines that are replaced by placeholders anyway. 

Tested when writing #354


terjekv (approved):
LGTM after being spoon-fed what was going on. :)

PR #356, opened 2024-12-04
Fix: Must check if uv is used to run manage.py

oyvindhagberg:


terjekv (approved):
Nice catch for backwards (non-uv) compatibility!

PR #357, opened 2024-12-04
Replace field models with custom types

pederhan:
Replaces models such as `HostT`, `IPAddressField` and `MacAddressField` with simpler custom types that Pydantic understands how to validate and serialize. It also makes them easier to use in the application, since we don't have to access the internal field of the model i.e. `HostT.hostname`, `IPAddressField.address`, etc. whenever we want to use it.

At runtime, the new `HostName` and `MacAddress` types are just string subclasses, while `IP_AddressT`, which we already use throughout the application, is a union of `ipaddress.IPv4Address` and `ipaddress.IPv6Address`. We can differentiate between the two IP types with the `.version` property.

## String types

New types:

- `mreg_cli.api.fields.MacAddress`
- `mreg_cli.api.fields.HostName`

These new types allow us to define functions that take in `HostName`, `MacAddress`, etc. instead of strings, so that we know we are operating on a valid value, while still having the ease of use of working with strings directly instead of through a proxy class.

## IP types

As mentioned in the introduction, we now operate on `IPv{4,6}Address` types directly, which we validate through `NetworkOrIP.parse(..., mode="ip")`, instead of having parallel interface for _just_ IP validation in the former `IPAddressField`.

## Benefits of this PR

By using the correct types directly in the model definitions, the models will serialize more cleanly. This is important want to tackle JSON output, since the new types serialize to basic types such as `str` instead of a nested dict containing a single key (`address`, `hostname`, etc.). It also removes a lot of custom field validators that were necessary to make these models work previously.

It also improves the ergonomics of these types in development, since, as mentioned, hostnames and mac addresses are just strings now (but with the benefit of being different from regular strings from the type checker's perspective).

### DIFF
--- a/mreg_cli/api/fields.py
+++ b/mreg_cli/api/fields.py
@@ -3,6 +3,11 @@
 The types validate to basic types like str, int, etc., but with additional
 validation added to them. The types are used in Pydantic models for consistent
 validation of common fields such as hostnames, MAC addresses, etc.
+
+Warning:
+Values constructed from these types should NOT be checked at runtime with isinstance()!
+Pydantic will always coerce these types to their schema types (str, int, etc.).
+
 """
 
 from __future__ import annotations
--- a/mreg_cli/api/fields.py
+++ b/mreg_cli/api/fields.py
@@ -3,15 +3,106 @@
 from __future__ import annotations
 
 import logging
+import re
 from typing import Annotated, Any
 
-from pydantic import AfterValidator, BeforeValidator
+from pydantic import AfterValidator, BeforeValidator, GetCoreSchemaHandler
+from pydantic_core import core_schema
 from pydantic_extra_types.mac_address import MacAddress as PydanticMacAddress
 
+from mreg_cli.config import MregCliConfig
 from mreg_cli.exceptions import InputFailure
 from mreg_cli.types import get_type_adapter
 
 logger = logging.getLogger(__name__)
+
+
+class HostName(str):
+    """Hostname string type."""
+
+    @classmethod
+    def parse(cls, obj: Any) -> HostName | None:
+        """Parse a MAC address from a string. Returns None if the MAC address is invalid.
+
+        :param obj: The object to parse.
+        :returns: The MAC address as a string or None if it is invalid.
+        """
+        try:
+            return cls.parse_or_raise(obj)
+        except InputFailure:
+            return None
+
+    @classmethod
+    def parse_or_raise(cls, obj: Any) -> HostName:
+        """Parse a hostname from a string. Returns the MAC address as a string.
+
+        :param obj: The object to parse.
+        :returns: The hostname as a string.
+        :raises ValueError: If the object is not a valid hostname.
+        """
+        try:
+            adapter = get_type_adapter(cls)
+            return cls(adapter.validate_python(obj))
+        except ValueError as e:
+            raise InputFailure(f"Invalid MAC address '{obj}'") from e
+
+    @staticmethod
+    def validate_hostname(value: str) -> str:
+        """Validate the hostname."""
+        value = value.lower()
+
+        if re.search(r"^(\*\.)?([a-z0-9_][a-z0-9\-]*\.?)+$", value) is None:
+            raise InputFailure(f"Invalid input for hostname: {value}")
+            # raise PydanticCustomError(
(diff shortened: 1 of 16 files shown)


mreg_cli/commands/host_submodules/a_aaaa.py
```diff
@@ -182,7 +183,8 @@ def _add_ip(
 
     mac = None
     if args.macaddress:
-        mac = MACAddressField.validate(args.macaddress)
+        # FIXME: should this raise exception if validation fails? Surely?
+        mac = MacAddress.parse(args.macaddress)
```
pederhan, line 187: This needs to be looked at

mreg_cli/commands/host_submodules/rr.py
```diff
@@ -95,7 +95,7 @@ def hinfo_add(args: argparse.Namespace) -> None:
     host = host.refetch()
 
     if host.hinfo and host.hinfo.cpu == args.cpu and host.hinfo.os == args.os:
-        OutputManager().add_ok(f"Added HINFO record for {host.name.hostname}.")
+        OutputManager().add_ok(f"Added HINFO record for {host.name}.")
```
pederhan, line 98:
```suggestion
        OutputManager().add_ok(f"Added HINFO record for {host.name!r}.")
```

Should probably be displayed in single quotes like most other places?

terjekv: Yes please.

terjekv: We should do a full pass and just do this.

Later. :)

mreg_cli/commands/host_submodules/a_aaaa.py
pederhan: I just changed this to `MacAddress.parse_or_raise`, because it makes no sense to silently fail here, even though we did that before this PR.

terjekv (approved):
I still find this weird, but it's better weird than the other weird.

mreg_cli/commands/host_submodules/core.py
```diff
@@ -16,13 +16,12 @@
 import argparse
 import re
 
+from mreg_cli.api.fields import HostName, MacAddress
```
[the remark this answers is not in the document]
pederhan: Yes, they can be moved to some `types` module if we want to. The thing that separates them from other type definitions is that they carry some validation logic that is specific to Pydantic, and they _are_ pretty much just used as type annotations for fields.

mreg_cli/api/models.py
```diff
@@ -2184,55 +2118,49 @@ class CNAME(FrozenModelWithTimestamps, WithHost, WithZone, WithTTL, APIMixin):
...
+    def get_by_host_and_name(cls, host: HostName | int, name: HostName) -> CNAME:
         """Get a CNAME record by host and name.
 
         :param host: The host to search for, either a hostname or an ID.
         :param name: The name to search for.
         :returns: The CNAME record if found, None otherwise.
         """
         target_hostname = None
-        if isinstance(host, HostT):
+        if isinstance(host, str):
```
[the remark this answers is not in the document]
pederhan: The only way to obtain a `HostName` (bar creating it manually, which we should never do), is through some validator function or via a model, so in essence we still have the same guarantees here about having a valid hostname, as long as the code passes type checking. The annotation remains `HostName`, but we just shouldn't check it at runtime.

I mean... We could prohibit _all_ instantiation of `HostName`, and _only_ use it as a Pydantic field type, but I have a feeling that would still cause just as many, if not more, problems in the long term.

mreg_cli/api/fields.py
```diff
@@ -1,48 +1,117 @@
...
-            raise InputFailure(f"Invalid MAC address '{value}'") from e
+            return cls.parse_or_raise(obj)
+        except InputFailure:
+            return None
 
     @classmethod
-    def parse_or_raise(cls, obj: Any) -> MacAddress:
-        """Parse a MAC address from a string. Returns the MAC address as a string.
+    def parse_or_raise(cls, obj: Any) -> HostName:
+        """Parse a hostname from a string. Returns the MAC address as a string.
```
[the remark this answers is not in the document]
pederhan: Careless copy-pasting.

PR #359, opened 2024-12-06
Validation error handling

pederhan:
This PR adds handling of Pydantic validation errors to the application, ensuring that such errors are printed and logged, and also prevents the application from crashing when such an exception is raised.

## Custom error messages

Instead of displaying the default verbose Pydantic validation error, we construct our own more concise version.

### With URL

When we have a validation error that happens as a result of failed validation of an API response, we show the request method and URL in the error message:

```
super@mreg-dev.uio.no> host info auspex
ERROR: Failed to validate Host response from GET https://mreg-dev.uio.no/api/v1/hosts/auspex.uio.no
  Input: foo
  Errors:
    Field: id
    Reason: Input should be a valid integer, unable to parse string as an integer
```

In this example, I simply patched the value of `"id"` in the response JSON to trigger the error:

```py
data["id"] = "foo"
```

### Without URL

If a validation error happens before we have made any requests in the current command invocation context, we omit the method and URL:

```
super@mreg-dev.uio.no> host info auspex
ERROR: Failed to validate Host
  Input: {'name': 'foo'}
  Errors:
    Field: created_at
    Reason: Field required

    Field: updated_at
    Reason: Field required

    Field: id
    Reason: Field required

    Field: ipaddresses
    Reason: Field required

    Field: contact
    Reason: Field required

    Field: comment
    Reason: Field required
```

In this example, I tried to construct a Host missing a number of required fields before sending any requests:

```py
    Host(name="foo") # exception is raised here
    for host in args.hosts:
        Host.get_by_any_means_or_raise(host, inform_as_cname=True).output(
            traverse_hostgroups=args.traverse_hostgroups
        )
```

## Context vars

To keep track of the last request method + url, this PR introduces 2 [`ContextVar`](https://docs.python.org/3.12/library/contextvars.html#contextvars.ContextVar) objects in `mreg_cli.utilities.api` named `last_request_url` and `last_request_method`.

The benefit of `ContextVar` over globals is that they are thread-safe and asyncio-compatible. If we introduce some sort of concurrency via threads or asyncio, we can be confident that threads or tasks cannot clobber each other's `last_request_url` and `last_request_method`.

### DIFF
--- a/mreg_cli/cli.py
+++ b/mreg_cli/cli.py
@@ -16,6 +16,7 @@ from typing import TYPE_CHECKING, Any, Protocol
 
 from prompt_toolkit import HTML, document, print_formatted_text
 from prompt_toolkit.completion import CompleteEvent, Completer, Completion
+from pydantic import ValidationError as PydanticValidationError
 
 # Import all the commands
 from mreg_cli.commands.dhcp import DHCPCommands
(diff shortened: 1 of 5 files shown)


terjekv (approved):
Very nice!

pederhan:
It occured to me that this should probably be implemented via an `ErrorBuilder`, similar to how we construct filter error messages. 

We can merge this for now, then consider if we want to refactor it to an error builder in the future. I think the application is due for a comprehensive exception handling PR sometime soon.

terjekv:
Yeah, we should move to ErrorBuilder and redo exception handling, but the poor Yak is sore enough for now. :)

PR #360, opened 2024-12-06
Disable uniqueness check by default

pederhan:
This PR disables the uniqueness check in `get_list_unique` and just returns the first result by default. The old behavior can be enabled with the new `expect_one_result` parameter for `get_list_unique`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Fixed
+
+- `dhcp assoc` command displaying unique constraint error instead of a proper error message when a MAC is already associated with an IP.
+
+### Changed
+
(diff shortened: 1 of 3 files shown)


terjekv (commented):
I'm still not sold on this. I think our issue is contextualising what we encounter multiples of more than the fact that we error on it. I need to think a bit on it.

terjekv (approved):
I still think we're seeing a problem with the approach combined with expectations, but this solution is better than what we have. I'm sold. Ish. :)

pederhan:
Closing for now, but we need to take another look at how we judge the validity of results. 

I personally don't believe we should be raising an exception so deep in the call stack for what is really a valid result from the API's perspective. 

Example of the call stack for `IPAddress.get_by_mac`:

- `IPAddress.get_by_mac`
- `APIMixin.get_by_field`
- `get_item_by_key_value` <- Called because `IPAddress.endpoint().requires_search_for_id() == False`
- `get_list_unique`
- `get_list_generic(..., expect_one_result=True)` <- Exception is raised here
- `get`
- `_request_wrapper`

Completely bricking a command because an endpoint returns _more_ data than we expect makes things very brittle. We can guard these calls in as many try/excepts as we want, but it won't unbrick the affected commands. It will just let the application inform users more eloquently about what we need to fix. 

To me, that is the crux of the issue; _is it_ something we need to fix? We allow creating these resources with `-force` and the API straight up allows it, so for us to assume every `?key=value` query can ever return a single result seems way too dangerous.

PR #361, opened 2024-12-08
Refactor MAC to IP association checking

pederhan:
Alternative to #360 to fix MAC to IP association checks.

This solves the same issue by fetching mac addresses via `get_list_by_field`, so we can inform the user about _all_ the IP addresses associated with the MAC address.

However, I would still say this is just a band-aid fix, and sets us up for similar bugs in the future.

### DIFF
--- a/ci/testsuite-result.json
+++ b/ci/testsuite-result.json
@@ -27485,7 +27485,7 @@
     "command_issued": "host add -ip 10.0.0.0/24 -macaddress aa:bb:cc:cc:bb:aa directfail",
     "ok": [],
     "warning": [
-      "MAC address aa:bb:cc:cc:bb:aa is already associated with 10.0.0.4"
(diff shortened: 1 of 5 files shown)


ci/testsuite-result.json
```diff
@@ -27485,7 +27485,7 @@
     "command_issued": "host add -ip 10.0.0.0/24 -macaddress aa:bb:cc:cc:bb:aa directfail",
     "ok": [],
     "warning": [
-      "MAC address aa:bb:cc:cc:bb:aa is already associated with 10.0.0.4"
+      "MAC address aa:bb:cc:cc:bb:aa is already associated with IP address 10.0.0.184, must force."
```
terjekv, line 27488: We should look at supporting `allocation_method: first`.

terjekv (approved):
LGTM

PR #364, opened 2024-12-12
Upgrade syntax with pyupgade in 3.11 mode

pederhan:
Also fixes linting errors found when running `ruff check --fix` (to remove redundant imports after upgrading syntax).

### DIFF
--- a/mreg_cli/api/history.py
+++ b/mreg_cli/api/history.py
@@ -77,7 +77,7 @@ class HistoryItem(BaseModel):
         """Clean up the timestamp for output."""
         return self.timestamp.strftime("%Y-%m-%d %H:%M:%S")
 
-    def msg(self, basename: str) -> str:
+    def msg(self, basename: str) -> str:  # noqa: ARG002 # arg might be used in the future
         """Attempt to make a history item human readable."""
         msg = ""
         action = self.action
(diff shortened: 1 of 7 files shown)


mreg_cli/api/history.py
```diff
@@ -77,7 +77,7 @@ def clean_timestamp(self) -> str:
         """Clean up the timestamp for output."""
         return self.timestamp.strftime("%Y-%m-%d %H:%M:%S")
 
-    def msg(self, basename: str) -> str:
+    def msg(self, basename: str) -> str:  # noqa: ARG002 # arg might be used in the future
```
terjekv, line 80: Is there a reason not to prefix with `_` rather than adding noqa to the line?

terjekv (approved):
LGTM

mreg_cli/api/history.py
pederhan: Mostly just a habit because I don't like messing with parameter names, which can cause issues when callers rely on them as named arguments (i.e. `msg(basename="foo")`). 

In this case, it's only called in a single place with where it's used with a positional argument, so we could just rename it, yeah.

terjekv: Tbh, if we are sending something as a named parameter and not using it, we have other issues. :) As for renaming, I end up using `Rename symbol` in VSCode or similar and that takes care of all the relevant refactoring unless we do massively weird stuff.

pederhan: Last time I tried refactoring tools in vscode it kind of sucked. I guess it works now, huh.

PR #366, opened 2024-12-16
Use typed parameters in a_aaaa helper functions

pederhan:
Refactors helper functions to take in discrete typed arguments. This improves IDE autocompletion and type checking inside helper functions.

Ultimately, this is a completely superficial "type safety"; we are the mercy of argparse and our own command parameter definitions. However, it's still an improvement of arbitrary `argparse.Namespace` attribute access and `Any` annotations.


terjekv (approved):
Our poor Yak. :)

PR #368, opened 2025-01-22
Add `help health` command

pederhan:
Adds a new `help health` command utilizing the new `/meta/health` endpoints. Shows uptime and LDAP status currently:

```
> help health
Health Information:
  Uptime: 0:42:34
  LDAP: OK
```

Written in a way that makes it easy to expand if new health endpoints are added in the future.

If the server doesn't implement the endpoints, the message shows `Unknown`:

```
> help health
Health Information:
  Uptime: Unknown
  LDAP: Unknown
```

## Expanded `APINotOk` exception

The `result_check()` function is mandatory for all HTTP request made through `_request_wrapper`, but prior to this PR, all non-2xx responses would cause the response to be lost, as the function would raise a `APINotOK` exception - discarding the response. This PR adds the response to the raised exception, so that callers catching `APINotOk` can still access it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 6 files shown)


terjekv (approved):
LGTM

PR #370, opened 2025-01-27
Flag value autocomplete

pederhan:
Autocomplete functionality for flag values such as networks, network policies, communities, etc.

### DIFF
--- a/mreg_cli/cli.py
+++ b/mreg_cli/cli.py
@@ -12,13 +12,16 @@ import os
 import shlex
 import sys
 from collections.abc import Generator
-from typing import TYPE_CHECKING, Any, Protocol
+from datetime import datetime
+from functools import lru_cache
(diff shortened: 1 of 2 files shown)


pederhan:
Too much has changed, should revisit this in the future. We need to be somewhat clever if we want to do this: background threads for fetching, caching, optimizing performance (what does it look like to try to display thousands of autocomplete values?), etc.

However, if I were to be realistic, I actually think current MREG probably can't handle the strain of multiple clients fetching this data constantly. If we had a separate endpoint that could deliver simple autocomplete values as strings (CIDR, hostnames, policy/atom names, label names, group names, etc.) and preferably cached for re-use by other clients, then we could provide this functionality pretty easily, I reckon.

terjekv:
I agree, we should ensure there is proper server side support for tailored, fast, and low-load data delivery before moving forward with this concept.

PR #371, opened 2025-01-28
Fix detection of reverse zones.

terjekv:
- Use the default way (the name ending in .arpa) to see if a zone is indeed a reverse zone.


pederhan (approved):
LGTM

PR #374, opened 2025-01-31
Make `short_desc` optional

pederhan:
This PR makes the `short_desc` parameter of `CommandRegistry.register_command()` optional, and defaults to the value of `description` if omitted. 

A lot of commands have identical `description` and `short_desc` arguments. The changes in this PR save developers some time and code duplication in such cases.


terjekv (approved):
LGTM

PR #375, opened 2025-03-04
Show exact override options for `host remove`

pederhan:
This PR improves the warning message shown in `host remove` when force with overrides is required. After printing all warnings, the exact force and override combination required is displayed.

Also fixes the `ipaddress` override, which prior to this PR was unusable due to the override argument processing only accepting `ipaddress`, while the actual override check only looked for `ipaddresses`.

### DIFF
--- a/ci/run_testsuite_and_record_V2.sh
+++ b/ci/run_testsuite_and_record_V2.sh
@@ -59,7 +59,7 @@ else
 fi
 
 # give mreg some time to create the database schema and start up
-sleep 5s
+sleep 5
 
 # create a superuser
 if docker exec -t mreg uv version >/dev/null; then
--- a/mreg_cli/commands/host_submodules/core.py
+++ b/mreg_cli/commands/host_submodules/core.py
@@ -255,9 +255,10 @@ def remove(args: argparse.Namespace) -> None:
         return False
 
     warnings: list[str] = []
+    overrides_required: set[str] = set()
     # Require force if host has any cnames.
     if host.cnames and not args.force:
-        warnings.append(f"  {len(host.cnames)} cnames, override with 'cname'")
+        warnings.append(f"  {len(host.cnames)} cnames")
         for cname in host.cnames:
             warnings.append(f"    - {cname.name}")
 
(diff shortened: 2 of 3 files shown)


ci/run_testsuite_and_record_V2.sh
```diff
@@ -59,7 +59,7 @@ else
 fi
 
 # give mreg some time to create the database schema and start up
-sleep 5s
+sleep 5
```
pederhan, line 62: Uhm, this snuck in, but we actually do need to remove the "s" for the script to work on MacOS.

mreg_cli/commands/host_submodules/core.py
```diff
@@ -255,9 +277,11 @@ def forced(override_required: str | None = None) -> bool:
...
 
     warnings: list[str] = []
+    overrides_required: set[str] = set()
     # Require force if host has any cnames.
-    if host.cnames and not args.force:
-        warnings.append(f"  {len(host.cnames)} cnames, override with 'cname'")
+    if host.cnames and not forced(Override.CNAME):
+        overrides_required.add(Override.CNAME)
+        warnings.append(f"  {len(host.cnames)} cnames")
         for cname in host.cnames:
```
pederhan, line 285: Require `cname` override or not? Can remove it if we want to treat it as a special case alongside the "multiple addresses on same VLAN" case, where we don't require an override, only force.

terjekv: Override on `cname` is probably overkill, but it is consistent and now self-documenting, so I think it's fine to have it.

ci/run_testsuite_and_record_V2.sh
terjekv: I think `s` implied most places, so this shouldn't hurt anything. 

Which are famous last words, but oh well.

mreg_cli/commands/host_submodules/core.py
```diff
@@ -201,6 +202,27 @@ def add(args: argparse.Namespace) -> None:
     host.output()
 
 
+class Override(str, Enum):
```
terjekv, line 205: Please and thank you! :)

terjekv (approved):
Very nice implementation. Thank you!

LGTM!

PR #376, opened 2025-03-20
Add support for expanded /hosts endpoint output

pederhan:
Adds support for the expanded /hosts endpoint output introduced in unioslo/mreg#572. Integration tests were performed against the container image produced here locally: https://github.com/unioslo/mreg/actions/runs/13944068027

The test suite will not pass here until we have merged unioslo/mreg#572.

### DIFF
--- a/mreg_cli/api/models.py
+++ b/mreg_cli/api/models.py
@@ -3484,38 +3484,9 @@ class Host(FrozenModelWithTimestamps, WithTTL, WithHistory, APIMixin):
 
         return False
 
-    def get_naptrs(self) -> list[NAPTR]:
-        """Return a list of NAPTR records."""
-        return NAPTR.get_list_by_field("host", self.id)
-
-    def get_srvs(self) -> list[Srv]:
-        """Return a list of SRV records."""
-        return Srv.get_list_by_field("host", self.id)
-
-    def get_sshfps(self) -> list[SSHFP]:
-        """Return a list of SSHFP records."""
-        return SSHFP.get_list_by_field("host", self.id)
-
     def get_roles(self) -> list[Role]:
         """List all roles for the host."""
         return Role.get_list_by_field("hosts", self.id)
-
-    def bacnet(self) -> BacnetID | None:
-        """Return the BacnetID for the host."""
-        if not self.bacnetid:
-            return None
-
-        return BacnetID.get_by_id(self.bacnetid)
-
-    def has_mx_with_priority(self, mx_arg: str, priority: int) -> MX | None:
-        """Check if the host has an MX record.
-
-        :param mx: The MX record to check for.
-        :param priority: The priority of the MX record.
-
-        :returns: True if the host has the MX record, False otherwise.
-        """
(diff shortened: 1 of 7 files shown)


mreg_cli/api/models.py
```diff
@@ -1281,7 +1281,7 @@ def output_atoms(self, _padding: int = 14) -> None:
             manager.add_line("No atom members")
 
     @classmethod
-    def output_multiple(cls, roles: list[Role], padding: int = 14) -> None:
+    def output_multiple(cls, roles: list[Role] | list[str], padding: int = 14) -> None:
```
pederhan, line 1284: It seemed silly _not_ to support `list[Role]` here, but this method is now more limited due to also having to accept lists of role names. Could consider refactoring into 2 methods, one for each type, but deemed it not necessary here (yet).

mreg_cli/utilities/api.py
```diff
@@ -295,7 +295,7 @@ def _request_wrapper(
         result.status_code == 500
         and (operation_type == "post" or operation_type == "patch")
         and params == {}
-        and data is not None
+        and data
```
pederhan, line 298: This snuck in while fixing pyright errors, but it's nice to finally silence that error, so we should keep it in.

terjekv: Works for me.

mreg_cli/api/models.py
```diff
@@ -3785,7 +3799,7 @@ def endpoint(cls) -> Endpoint:
 
     @classmethod
     def output_multiple(
-        cls, hostgroups: list[HostGroup], padding: int = 14, multiline: bool = False
+        cls, hostgroups: list[HostGroup] | list[str], padding: int = 14, multiline: bool = False
```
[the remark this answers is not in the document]
terjekv: Yep.

terjekv:
Just out of curiosity and as an answer to "why", can you compare the runtimes of `host info` for a host with and without the changes?

pederhan:
> Just out of curiosity and as an answer to "why", can you compare the runtimes of `host info` for a host with and without the changes?

Take this with a grain of salt:

Benchmarked with hyperfine using:

```
hyperfine --warmup 5 -n 20 "mreg-cli --url https://<redacted> --domain uio.no --user <redacted> host info testhost"
```

On this branch (36a8cb8)

```
Benchmark 1: 36a8cb8
  Time (mean ± σ):     907.8 ms ± 177.7 ms    [User: 279.7 ms, System: 49.3 ms]
  Range (min … max):   777.6 ms … 1516.9 ms    20 runs
```
  
  
On master:
  
```
Benchmark 1: master
  Time (mean ± σ):      1.393 s ±  0.052 s    [User: 0.290 s, System: 0.051 s]
  Range (min … max):    1.322 s …  1.508 s    20 runs
```

terjekv:
It'd be interesting to see the start + logout, but it looks like a big improvement. Good stuffs, thank you!

terjekv (approved):
It was discussed to retain support for older servers where the host data returned is more limited, but it was found to be too much code to solve a problem that has the simple solution of upgrading. 

LGTM.

PR #377, opened 2025-03-20
Re-raise `LoginFailedError` in `prompt_for_password_and_login`

pederhan:
If we have already raised a `LoginFailedError` in `auth_and_update_token()`, we should not obscure it with another error with a less precise message.


terjekv (approved):
LGTM

PR #379, opened 2025-03-26
CI: Show summary of diffs after completion

pederhan:
This PR adds summaries of resolved and unresolved diffs after completion in `ci/diff.py` that looks like this:

image

It required a moderately large refactoring of the data structures we pass around in the script to implement this feature, but the result is (hopefully) improved readability and better type annotations.


terjekv (approved):
LGTM. :)

PR #380, opened 2025-04-09
Keep new test suite log when not in CI

pederhan:
When a manual diff resolution is required (such as when adding/removing commands), we need to extract the new test suite results from the container to do so. This PR ensures the `new_testsuite_log.json` file is always copied over from the container.


terjekv (approved):
LGTM

PR #382, opened 2025-04-10
Require force for `host {a,aaaa}_remove` with CNAME arg

pederhan:
This PR adds a `-force` requirement when `host {a,aaaa}_remove` is called with a CNAME argument.

We should probably update the test suite with tests for this, but I would rather not make changes to the test suite while #369 is in progress. The issue #385 has been created as a reminder.

### DIFF
--- a/mreg_cli/commands/host_submodules/a_aaaa.py
+++ b/mreg_cli/commands/host_submodules/a_aaaa.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 import argparse
 
 from mreg_cli.api.fields import HostName, MacAddress
-from mreg_cli.api.models import Host, HostList, IPAddress, Network, NetworkOrIP
+from mreg_cli.api.models import CNAME, Host, HostList, IPAddress, Network, NetworkOrIP
 from mreg_cli.commands.host import registry as command_registry
 from mreg_cli.exceptions import (
     DeleteError,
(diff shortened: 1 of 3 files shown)


mreg_cli/commands/host_submodules/a_aaaa.py
```diff
@@ -142,6 +143,14 @@ def _ip_remove(name: str, ipaddr: str, ipversion: IP_Version) -> None:
...
         raise EntityNotFound(f"Host {host} does not have IP {ip}")
 
+    # Check if we fetched the host via a CNAME.
+    # Ensure arg name ends with configured domain
+    # (e.g. "foo" -> "foo.example.com")
+    name_hostname = HostName.parse_or_raise(name)
+    for cname in host.cnames:
+        if not force and name_hostname == cname.name.casefold():
+            raise ForceMissing(f"{name_hostname} is a CNAME for {host.name}, must force.")
+
```
pederhan, line 153: This solution isn't great, but it was the best way I could find that avoids additional API calls.

With API calls, we could do this:

```suggestion
    if not force and host.cnames:
        name_hostname = HostName.parse_or_raise(name)
        cname = CNAME.get_by_field("name", name_hostname)
        if cname:
            raise ForceMissing(f"{cname.name} is a CNAME for {host.name}, must force.")
```

PR #383, opened 2025-04-14
Update uv.lock to revision 1

pederhan:
Chore: Update uv.lock to revision 1

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,4 +1,5 @@
 version = 1
+revision = 1
 requires-python = ">=3.11"
 resolution-markers = [
     "python_full_version < '3.13'",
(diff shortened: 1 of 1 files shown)


terjekv:
> Newer versions of Uv will automatically these changes when running `uv lock`. Getting this chore out of

I'm... not sure I follow this sentence? :)

pederhan:
Oops. 

> Newer versions of uv will automatically __make__ these changes. 

That is to say, when a recent version of uv (>=0.6.0) modifies the lockfile, it will add `revision` along with whatever changes that revision bring with it.

terjekv:
Ah, that makes more sense. :)

terjekv (approved):
Probably fine? 👍

pederhan:
Who knows. Revision 2 has already been released since I made this PR, so I might try to force update the lockfile so we can jump directly to revision 2 with this PR.

PR #384, opened 2025-04-14
Format DRF error message on failed login

pederhan:
**NOTE:** This PR should probably be scrapped in favor of #386.

----

#377 changed the login error handling to propagate the original Django error. However, we did not account for the fact that DRF returns a JSON response with a single `"detail"` field. This resulted in the following error message:

```
Connecting to https://mreg.example.com
Password for pederhan: ******
Login failed: {'detail': 'Incorrect authentication credentials.'}
```

This PR attempts to extract the `"detail"` value before constructing the exception:

```
Connecting to https://mreg.example.com
Password for pederhan: ******
Login failed: Incorrect authentication credentials.
```

## Refactored fallback for errors without `"detail"` field

Currently, we have a specific check for 400 responses including `non_field_errors`:

https://github.com/unioslo/mreg-cli/blob/9f14d178194077f5852d673e36e23615d482e908/mreg_cli/utilities/api.py#L197-L201

This has the side-effect of raising no exception at all if the response doesn't contain `non_field_errors`, where the application will instead continue on trying to update the token, even though the response does contain a token.

This PR adds more generalized handling of this class of errors, so that we cannot fall through with no raised exception:

https://github.com/unioslo/mreg-cli/blob/095f35ec66f1e636b3802719a3dd91ece97ea492/mreg_cli/utilities/api.py#L194-L205

### DIFF
--- a/mreg_cli/utilities/api.py
+++ b/mreg_cli/utilities/api.py
@@ -181,6 +181,8 @@ def prompt_for_password_and_try_update_token() -> None:
 
 def auth_and_update_token(username: str, password: str) -> None:
     """Perform the actual token update."""
+    from mreg_cli.api.models import ErrorResponse  # avoid circular import
+
     tokenurl = urljoin(MregCliConfig().get_url(), "/api/token-auth/")
     logger.info("Updating token for %s @ %s", username, tokenurl)
     try:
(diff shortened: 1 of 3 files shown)


mreg_cli/utilities/api.py
```diff
@@ -189,16 +191,19 @@ def auth_and_update_token(username: str, password: str) -> None:
...
-            res = result.text
-        if result.status_code == 400:
-            if "non_field_errors" in res:
-                raise LoginFailedError("Invalid username/password")
-        else:
-            raise LoginFailedError(res)
+        err = ErrorResponse.to_string(result)
+        # We don't have an error response with a "detail" field
+        if not err:
+            body = result.text
```
pederhan, line 199: `requests.Response.text` is a property that is computed every time we access it (no explicit caching involved). Binding the result to a variable ensures we only perform that computation once.

pederhan:
Closing this in favor of #386.

PR #387, opened 2025-05-19
Support last_login and token information about users.

terjekv:
This requires https://github.com/unioslo/mreg/pull/579 to expose the data. For older servers both "Last login" and the token information will be None. 

Example output for someone that has logged in with an active session:

```
Username: user
Last login: 2025-05-13T10:05:17.553790+02:00
Token:
  Valid: True
  Created: 2025-05-13T10:05:17.526321+02:00
  Expires: 2025-05-13T18:05:17.526321+02:00
  Last used: 2025-05-13T10:40:56.790783+02:00
  Lifespan left: 7:24:20.704934
[...]
```


pederhan (approved):
LGTM!

PR #388, opened 2025-05-22
Add `host add -force` for broadcast and network addresses

pederhan:
Allows using network or broadcast address as `-ip` argument in `host add` if `-force` is used and the user has permission to do so (NYI on the server).

### DIFF
--- a/mreg_cli/commands/host_submodules/core.py
+++ b/mreg_cli/commands/host_submodules/core.py
@@ -141,13 +141,13 @@ def add(args: argparse.Namespace) -> None:
             try:
                 network = Network.get_by_ip(ipaddr)
                 if network:
-                    if ipaddr == network.network_address:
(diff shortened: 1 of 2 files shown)


mreg_cli/commands/host_submodules/core.py
```diff
@@ -141,13 +141,13 @@ def add(args: argparse.Namespace) -> None:
             try:
                 network = Network.get_by_ip(ipaddr)
                 if network:
```
pederhan, line 143:
```suggestion
                if network and not force:
```

I considered putting the force check here instead, but I think the more explicit approach I took with force checks for each case is more robust, since we may want to add more checks in the future that are not dependent on force status.

pederhan:
We may want to add tests for this feature once we have implemented it server side.

PR #389, opened 2025-06-05
Add support for fetching multiple hosts with the same IP

pederhan:
## Issue

Multiple hosts can be assigned the same IP. When multiple hosts have the same IP, using `host info` with an IP argument stops working.

```
> host info 123.123.123.123
Multiple hosts found with IP address 123.123.123.123.
```

## Solution

Always assume we can fetch more than one host in `host info` by adding a new `get_list_by_any_means_or_raise()` method.

## Further work / questions

This PR only fixes `host info`,  but `Host.get_by_any_means_or_raise()` is referenced _57_ times if we are to trust VS Code. I think it's perhaps unrealistic to go back and rewrite all fetching of hosts to assume we always have a list of hosts, as opposed to treating `host info` as an outlier in this regard.

### DIFF
--- a/mreg_cli/api/models.py
+++ b/mreg_cli/api/models.py
@@ -3019,6 +3019,22 @@ class Host(FrozenModelWithTimestamps, WithTTL, WithHistory, APIMixin):
         return Endpoint.Hosts
 
     @classmethod
+    def get_list_by_ip(cls, ip: IP_AddressT, inform_as_ptr: bool = True) -> list[Self]:
+        """Get a list of hosts by IP address.
+
+        :param ip: The IP address to search for.
+        :param check_ptr: If True, check for PTR overrides as well.
+        :returns: A list of Host objects.
+        """
+        hosts = cls.get_list_by_field("ipaddresses__ipaddress", str(ip))
+        if not hosts:
+            hosts = cls.get_list_by_field("ptr_overrides__ipaddress", str(ip))
+            if hosts and inform_as_ptr:
+                for host in hosts:
+                    OutputManager().add_line(f"{ip} is a PTR override for {host.name}")
+        return hosts
+
+    @classmethod
     def get_by_ip(cls, ip: IP_AddressT, inform_as_ptr: bool = True) -> Host | None:
         """Get a host by IP address.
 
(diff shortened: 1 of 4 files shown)


mreg_cli/api/models.py
```diff
@@ -3157,6 +3194,70 @@ def get_by_any_means(
...
+        :param ok404: If True, don't raise a EntityNotFound if the host is not found.
+        :param inform_as_cname: If True, inform the user if the host is a CNAME.
+        :param inform_as_ptr: If True, inform the user if the host is a PTR override.
+
+        :raises EntityNotFound: If we don't find the host and `ok404` is False.
+
+        :returns: A Host object if the host was found, otherwise None.
+        """
+        if identifier.isdigit():
+            return cls.get_list_by_id(int(identifier))
```
pederhan, line 3235: I was thinking that if we have an ID, we might be able to assume we should just do the following:

```suggestion
        if identifier.isdigit():
            return [cls.get_by_id(int(identifier))]
```

Since an ID shouldn't point to multiple hosts?

mreg_cli/api/models.py
```diff
@@ -3157,6 +3194,70 @@ def get_by_any_means(
...
+            return cls.get_list_by_ip(ip, inform_as_ptr=inform_as_ptr)
+
+        if mac := MacAddress.parse(identifier):
+            return cls.get_list_by_mac_or_raise(mac)
+
+        # Let us try to find the host by name...
+        identifier = HostName.parse_or_raise(identifier)
+
+        if hosts := cls.get_list_by_field("name", identifier):
+            return hosts
```
pederhan, line 3247: Similarly here and with CNAMES, wherein a name or ID should point to a _single_ host, not multiple, surely?

terjekv: Yeah, hostnames can't be shared. Neither can CNAMEs. They can have multiple IPs and such, but they're unique and distinct host/cname objects.

mreg_cli/api/models.py
```diff
@@ -3263,7 +3263,7 @@ def get_list_by_any_means(
         # If we found a CNAME, get the host it points to. We're not interested in the
         # CNAME itself.
         if cname is not None:
-            host = cls.get_list_by_id(cname.host)
+            hosts = cls.get_list_by_id(cname.host)
```
pederhan, line 3266: Oddly, this did not change the test results. Perhaps missing coverage in tests?

terjekv: > Perhaps missing coverage in tests?

😱 Can't be. Ever.

terjekv (approved):
Needed, but we should probably redesign bits of the internal APIl.

PR #396, opened 2025-08-26
Add simple disk-based caching

pederhan:
This PR adds simple disk-based caching using [diskcache](https://pypi.org/project/diskcache/). 

## Implementation

It is implemented as a transparent layer over the `mreg_cli.utils.api.get()` method, with an expiration time of 300 seconds. Results from `get()` are memoized and stored in the cache for a maximum of 300 seconds. Each time the user performs a write operation (i.e. POST, PUT, PATCH, DELETE), the cache is cleared. There is no granularity in the way items are stored in cache; all items are flushed if the user performs a write operation.

## Robustness

In case the cache cannot be initialized (lack of write privileges, sqlite errors, etc.), a dummy cache is instantiated, whose `memoize()` decorator simply returns the unmodified `get()` function. The most common cache methods (currently unused), are implemented as dummy methods with deterministic behavior compatible with the corresponding `diskcache.Cache` methods.

## Notes

Due to the relatively low frequency of writes compared to reads we witness, I deemed it acceptable to implement the cache in this all-or-nothing way. Trying to tag, set and delete cache items based on API operations would be a huge undertaking, while in practice yielding very few improvements over the current implementation. It would very likely also make the API code entangled with the cache code in a way that the current decorator-based approach does not.

### DIFF
--- a/mreg_cli/utilities/api.py
+++ b/mreg_cli/utilities/api.py
@@ -23,7 +23,7 @@ from requests import Response
 
 from mreg_cli.__about__ import __version__
 from mreg_cli.api.errors import parse_mreg_error
-from mreg_cli.cache import cache
+from mreg_cli.cache import get_cache
 from mreg_cli.config import MregCliConfig
 from mreg_cli.exceptions import (
     APINotOk,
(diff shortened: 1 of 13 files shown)


mreg_cli/utilities/api.py
```diff
@@ -591,20 +594,41 @@ def get_typed(
...
+        try:
+            cache.evict("api")
+        except Exception as e:
+            logger.error(f"Error clearing cache: {e}")
+        return result
+
+    return wrapper
+
+
+@clear_cache
```
terjekv, line 615: Will we ever need multiple tagged caches? I can't see a reason for it, but maybe passing the tag is cleaner? Or name it `@clear_entire_cache`? Dunno.

mreg_cli/utilities/api.py
```diff
@@ -333,6 +335,7 @@ def get(path: str, params: QueryParams | None = ..., *, ok404: bool) -> Response
 def get(path: str, params: QueryParams | None = ...) -> Response: ...
 
 
+@cache.memoize(expire=300, tag="api")
```
terjekv, line 338: Should we possibly only cache calls where we get 2XX from the server?

terjekv (approved):
Looks like a great improvement. We should debate cache time, but 5 minutes is probably acceptable.

mreg_cli/utilities/api.py
pederhan: Don't we raise an exception for non-2xx responses? In that case the memoized function should never get to the point at which it actually writes to the cache.

https://github.com/grantjenks/python-diskcache/blob/ebfa37cd99d7ef716ec452ad8af4b4276a8e2233/diskcache/core.py#L1874-L1877

The library just calls the decorated function with the given arguments, so if an exception is raised, it's propagated up the stack before the results are cached.

PR #399, opened 2025-09-09
Rewrite config as Pydantic model

pederhan:
This PR rewrites the MregCliConfig singleton as a Pydantic model, which gives us validation, autocomplete, and more flexibility in adding/removing sources to read config options from.

WIP


mreg_cli/dirs.py
pederhan, line 1: This module ended up becoming very ugly in order to support users without a home directory, and that rot spread to  `fs.py` and `config.py` as well:

https://github.com/unioslo/mreg-cli/blob/be098173adc656b1259f76933af1396c54ad8219/mreg_cli/utilities/fs.py#L39-L49

https://github.com/unioslo/mreg-cli/blob/be098173adc656b1259f76933af1396c54ad8219/mreg_cli/config.py#L121-L138

PR #404, opened 2025-10-17
Add `@disable_cache` decorator, don't cache `help` commands

pederhan:
This PR adds a `@disable_cache` decorator which can be added to functions to temporarily disable the program-wide cache when calling the function. The decorator has been applied to all `help` commands that make API calls to ensure these always return fresh data.

## Implementation

It works by unpatching all patched functions (but without clearing the cache!), and then re-patching the functions once the decorated function has executed. This was the simplest implementation that I could come up with, as diskcache does not expose an API for toggling caching on/off for memoized functions. However, the good news is that this implementation has _very_ minimal overhead:

```
super@mreg-dev.uio.no> help health
Took 0.080041 ms to disable cache
Health Information:
  Uptime: 18 days, 3:06:56
  LDAP: OK
super@mreg-dev.uio.no> help health
Took 0.031417 ms to disable cache
Health Information:
  Uptime: 18 days, 3:07:04
  LDAP: OK
super@mreg-dev.uio.no> help health
Took 0.042958 ms to disable cache
Health Information:
  Uptime: 18 days, 3:07:06
  LDAP: OK
```

<details>
  <summary>Modified code with <code>time.perf_counter_ns</code></summary>
  
  ```python
def disable_cache(func: Callable[P, T]) -> Callable[P, T]:
    """Disable cache for the duration of the decorated function."""

    @functools.wraps(func)
    def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
        cache = get_cache()
        try:
            start_time = time.perf_counter_ns()
            cache.disable(clear=False)
            stop_time = time.perf_counter_ns()
            print(f"Took {(stop_time - start_time) / 1_000_000} ms to disable cache")
            return func(*args, **kwargs)
        finally:
            cache.enable()

    return wrapper
```

</details>

### DIFF
--- a/mreg_cli/cache.py
+++ b/mreg_cli/cache.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import functools
 import logging
 from dataclasses import dataclass
 from types import ModuleType
@@ -243,10 +244,11 @@ class MregCliCache:
         self._patch_functions()
         logger.info("Cache enabled")
 
-    def disable(self) -> None:
+    def disable(self, *, clear: bool = True) -> None:
         """Disable caching by restoring patched functions and clearing the cache."""
         self._unpatch_functions()
-        self.clear()
+        if clear:
+            self.clear()
         logger.info("Cache disabled")
 
     def clear(self) -> None:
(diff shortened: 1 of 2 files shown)


terjekv (approved):
Hah! :D

PR #409, opened 2025-11-04
Add `policy list_hosts` `-exclude` option

pederhan:
Adds `-exclude` option for `policy list_hosts`, which allows for further filtering of hosts that _do not_ have the given role(s)

**Exclude a single role**

All hosts with the `foo` role but not the 'bar' role:

```
policy list_hosts foo -exclude bar
```

**Exclude multiple roles (multiple arguments)**

All hosts with the `foo` role but not the `bar` & `baz` roles:

```
policy list_hosts foo -exclude bar baz
```

**Exclude multiple roles (regex)**

All hosts with the `foo` role but not any roles starting with `ba`:

```
policy list_hosts foo -exclude "ba.+"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `policy list_hosts -exclude EXCLUDEROLE [EXCLUDEROLE ...]` option to exclude hosts that have the given role(s) when listing hosts with a role. Supports multiple arguments and regular expressions.
 - Environment variable support for all config file options.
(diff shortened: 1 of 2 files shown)


pederhan:
I have written tests for this feature here: https://github.com/unioslo/mreg-cli/compare/policy-list-hosts-exclude...policy-list-hosts-exclude-with-tests

## Changes

image

Summary of changes attached as a file.
[diff.txt](https://github.com/user-attachments/files/23357992/diff.txt)

## What to do

The branch adds _a lot_ of test suite changes, and causes 115 commands to change because we add a host and increment the last host ID in the db. We can do one of the following:

- Revise the tests to avoid incrementing host ID in the db (don't add a new host)
- Accept the changes as-is
- Not add tests at all

PR #455, opened 2026-04-16
Set user agent in mreg client init

### LINKED ISSUE #445: User agent configuration gone after #432
I forgot to add user agent configuration to mreg-api (and thus to mreg-cli in #432). It has been added back in mreg-api here: https://github.com/unioslo/mreg-api/pull/10

Will need to specify custom user agent when configuring client in mreg-cli.

pederhan:
Fixes #445


terjekv (approved):
LGTM!

PR #458, opened 2026-04-23
Fix `naptr_remove` inner loop break bug

pederhan:
Also refactors filtering as a whole so that fields that may be blank don't have to be specified. Now skips filtering on `flag`, `regex`, `service` unless specified.

Intentionally matching blank fields can be done via `-flag "" -regex "" -service ""`


terjekv (approved):
LGTM!

PR #459, opened 2026-04-24
CI: Increase health check start period

pederhan:
After switching to Docker VMM for virtualization on my M1 mac, the test suite started terminating itself before it could begin, because of increased startup time of containers. This PR adds a longer grace-period before we start measuring retries for the health check.


terjekv (approved):
LGTM

PR #461, opened 2026-05-21
Fix host add contact argument ordering

terjekv:
Fixes the regression where hostnames have to come first in the list due to the use of `nargs="+"` by making `-contact` a repeatable flag.


pederhan (approved):
LGTM

PR #465, opened 2026-07-04
Upgrade mreg api to 0.3.0

pederhan:
Uses new non-singleton client with dotted manager attribute access. 

Adds a `mreg_cli.client.get_client` function for accessing a globally configured client object shared by commands.

----

Numerous API calls have been removed from the testsuite because of more consistent host fetching (uses `/hosts/<name>` instead of `/hosts/?id=<id>`). This is a positive thing in practice, but it highlights the need to potentially disable caching in tests, because caching (and cache invalidation) makes it harder to see differences in API calls when a command suddenly hits the cached data set by a previous command. When an API call disappears from the test suite log, we can't verify at a glance whether the command didn't send the request or if it hit the cache.

### DIFF
--- a/mreg_cli/commands/network.py
+++ b/mreg_cli/commands/network.py
@@ -1368,24 +1368,81 @@ def community_set_description(args: argparse.Namespace) -> None:
     OutputManager().add_ok(f"Set new description for community {community!r}")
 
 
-def _check_host_ip(host: Host, ip: str | None) -> IPAddress:
-    """Ensure host has an IP that can be added to/removed from a community."""
+def _get_host_ip_to_add(host: Host, ip: str | None) -> IPAddress:
+    """Get the IP address to add to a community for a host.
+
+    Requires `ip` to be specified if the host has multiple IP addresses.
+
+    :param host: Host object
+    :param ip: Optional IP address string
+    :return: IPAddress object
+
+    :raises EntityNotFound: If the host has no IP addresses.
+    :raises EntityNotFound: If the given IP address is not associated with the host.
+    :raises InputFailure: If the host has multiple IP addresses and no IP is specified.
+    """
     if not host.ipaddresses:
         raise EntityNotFound(f"Host {host.name!r} is not associated with any networks.")
-    elif not ip and len(host.ipaddresses) > 1:
+
+    if not ip and len(host.ipaddresses) > 1:
         raise InputFailure(
             f"Host {host.name!r} is associated with multiple IP addresses. Must specify IP."
         )
 
+    # Disambiguate or choose the first IP address if only one is present
     if ip:
-        ip_t = NetworkOrIP.parse_or_raise(ip, mode="ip")
-        ipaddr = host.get_ip(ip_t)
-        if not ipaddr:
-            raise EntityNotFound(f"Host {host.name!r} is not associated with IP {ip_t}")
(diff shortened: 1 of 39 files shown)


mreg_cli/commands/policy.py
```diff
@@ -57,18 +54,19 @@ def atom_create(args: argparse.Namespace) -> None:
...
     # Check if atom with that name already exists
-    Atom.get_by_name_and_raise(name)
+    client.atom.ensure_absent(name)
 
     params = {"name": name, "description": description}
     if created:
         params["create_date"] = created
 
-    Atom.create(params)
+    client.atom.create(name=params["name"], description=params.get("description", ""))
```
terjekv: Above we set create_date if created, but this isn't propagated anywhere in the new interface, is that intended?

mreg_cli/commands/network.py
```diff
@@ -1367,14 +1436,27 @@ def community_host_remove(args: argparse.Namespace) -> None:
...
+
+    com = h.get_community(community, ipaddr)
+    if not com:
+        msg = f"Community {community!r}"
+        if ipaddr:
+            msg += f" for IP address {ipaddr}"
+        raise EntityNotFound(f"{msg} not found.")
+
+    com = client.network.community.get_by_name(community, net)
+    client.network.community.remove_host(com, net, h)
```
terjekv: Here we also don't send ip like we used to.

pederhan: We seem to have never passed the IP address to the API in this command (both before and after the introduction of `mreg-api`):

https://github.com/unioslo/mreg-api/blob/ed66fe457de067138fce2dd7dcd65049402b2aac/mreg_api/models/models.py#L2120-L2142

https://github.com/unioslo/mreg-cli/blob/12954176aa8a38d01fd2cc9b620c591878c130b9/mreg_cli/api/models.py#L2248-L2264

LLM refactor of that method simply decided to omit the param instead of warning that we forgot to pass it to the API. 🤪

terjekv: I mean, the LLM is not *wrong*... "You seem to not use this stuff, why are you keeping it around?"

Although, with that logic, it should just go "why do you want to talk to mreg at all? Have you seen the stuff that server does?" 😱

pederhan: Fixed in https://github.com/unioslo/mreg-api/pull/18/changes/b4873acd8bad998325c429537c3e1e7eb5709324

mreg_cli/commands/policy.py
pederhan: Executive decision here: deprecate option for setting the date manually.

https://github.com/unioslo/mreg-cli/blob/c4adc4f9a29f799e34b5cf1bab3b89e0f33e37e2/mreg_cli/commands/policy.py#L42-L56

https://github.com/unioslo/mreg-cli/blob/8ff7c2f6cfe668d0bda0e5f46f3dad43abcbd4a6/mreg_cli/commands/policy.py#L93-L102

pederhan:
Issues to create after this is merged:

- Subclass `MregClient` and add `mreg_cli.utilities.resolution.resolve_*` functions as methods on it.
- Remove references to `fetch_after_create` (POST now returns created object as of mreg v1.6.0).

terjekv (approved):
LGTM

